PR #14, opened 2026-04-12
fix: include character description in prompt

willie:
## Summary
- The character card `description` field (physical traits, background, appearance) was never included in the LLM prompt
- Characters imported from SillyTavern/Backyard AI typically put their core definition in `description`, not `personality`, so they were losing key attributes during generation
- Fix prepends `description` to the persona block via `_getEffectivePersonality()`, which feeds all prompt assembly paths (single chat, group chat, impersonate, story pipeline)

## Test plan
- [ ] Load a character card where traits are defined in `description` (not `personality`) — verify the model now reflects those traits
- [ ] Load a character with both `description` and `personality` filled — verify both appear in the persona block
- [ ] Load a character with empty `description` — verify no blank lines or changes in behavior
- [ ] Test with character evolution enabled — verify evolved traits still append correctly


willie:
Duplicate of #15 — closing this one since the dev branch has unrelated upstream commits mixed in.

PR #33, opened 2026-05-19
Feat/pseudo remote

S-A-M-F:
Added a new backend option (local koboldcpp but through OpenAI API - fixes some issues I'm unable to tackle so far), some architecture changes and QoL. Details in commits descriptions.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -77,6 +77,18 @@ jobs:
       - name: Install dependencies
         run: flutter pub get
 
+      - name: Disable inappwebview Linux plugin (uses external browser fallback)
+        if: matrix.platform == 'linux'
+        run: |
+          echo "" >> pubspec.yaml
+          echo "dependency_overrides:" >> pubspec.yaml
+          echo "  flutter_inappwebview_linux:" >> pubspec.yaml
+          echo "    path: ci/flutter_inappwebview_linux_stub" >> pubspec.yaml
+          flutter pub get
+
+      - name: Setup Rust Toolchain
+        uses: dtolnay/rust-toolchain@stable
+
       - name: Build Application
         if: matrix.platform != 'linux'
         run: flutter build ${{ matrix.platform }} --release
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -5,7 +5,6 @@ env:
 
 on:
   schedule:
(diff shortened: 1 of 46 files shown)


linux4life1:
Thanks for the PR! I really appreciate the work you put into this.

The core idea (running local KoboldCPP through the OpenAI-compatible path so local users get more reliable Realism Engine behavior) is solid, and it’s clear you spent real time on it — especially the process management, log handling, and the light mode scaffolding work. That kind of hands-on iteration is genuinely valued.

A couple of things came up when we tested the merge:

### 1. Branch target
The PR was originally opened against `main`. Per the project guidelines, all PRs should target `dev` (main is reserved for stable releases only). I went ahead and retargeted it to `dev` for you, but just wanted to flag it for future PRs.

### 2. Crash on startup
When we merged the PR into the latest `dev` and ran the app, we hit this error during the first build:

> `setState() or markNeedsBuild() called during build.`

The crash is triggered here:

```dart
// main.dart:794
Provider.of<StorageService>(context, listen: false).isDark = appState.darkMode;
```

This line lives inside the `Consumer<AppState>` in `_MyAppState.build`. When it runs, the `isDark` setter in `StorageService` calls `notifyListeners()`, which tries to mark widgets as needing rebuild *while the framework is still building*. Flutter doesn’t allow this.

It looks like this path became active (or was introduced) in the light mode scaffolding commit. The rest of the PR merged cleanly and tests pass, but this one spot is preventing the app from even starting.

---

Would you be up for taking another pass at it?

- Fixing the `isDark` sync so it doesn’t notify during build (a `postFrameCallback` or moving the logic out of the build method usually works)
- Making sure any future updates are based on `dev`

Happy to help review the fix once it’s ready, or discuss the best way to handle the dark/light sync if you want a second set of eyes.

Again, really appreciate you putting the time into this — the Pseudo-Remote direction is interesting and the effort shows. Looking forward to getting it in once we smooth out this startup crash.

linux4life1 (changes requested):
Thanks for the contribution and the effort put into this.

I've left a detailed comment on the PR with the issues that came up during testing:

- The PR was originally targeting `main` (should target `dev` per project guidelines).
- A crash on app startup: `setState() or markNeedsBuild() called during build`, caused by the dark mode sync logic in `main.dart` calling `StorageService.isDark` (which notifies listeners) during the build phase.

Requesting changes to address these points. The rest of the PR merged cleanly and the core idea is interesting.

Happy to re-review once the issues are fixed!

PR #45, opened 2026-06-06
test: realign stale assertions to green Rawhide CI

MisterLotto:
## Summary

Rawhide CI is red on every push. Of the 6 failing tests, **4 were stale assertions** — tests asserting old behavior that production code intentionally changed, not real regressions. This PR realigns those 4. Test-only change; no production code touched.

## What was stale

- `avatar_repository_test.dart` — asserted Drift `schemaVersion == 29`; the schema is now **32** (three migrations landed without the test being updated). Bumped to 32.
- `chat_message_test.dart` (3 tests) — the `ChatMessage` constructor now clamps an out-of-range `swipeIndex` to 0 (guards against corrupted DB rows, `chat_service.dart:188`). The tests predated the clamp:
  - "swipeIndex out of range throws RangeError" → rewritten to assert the new safe contract (clamps to 0, resolves to `swipes[0]`, no throw).
  - "activeMetadata at current index" / "pads with nulls" → given a `swipes` list long enough that their `swipeIndex` stays valid, preserving the original test intent rather than weakening assertions.

## ⚠️ CI will still show 2 failures after this merge — expected

These two are **pre-existing on Rawhide** (not introduced here) and are **out of scope** for this PR. They live in the parity-sensitive Realism Engine and may be real regressions needing code fixes, so they're being tracked and fixed separately:

- `chat_service_realism_engine_test.dart` — V2.5 seeding (`expected 55, got 110`)
- `chat_service_realism_engine_test.dart` — one-shot dynamic deltas (`expected true, got false`)

Splitting keeps this safe, mechanical fix from being held hostage to that investigation.

## Verification
- `flutter test test/services/chat_message_test.dart test/services/avatar_repository_test.dart` → 50 pass
- `flutter analyze` → clean

### DIFF
--- a/.claude/changelog.md
+++ b/.claude/changelog.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## 2026-06-06 (Realign stale test assertions to green Rawhide CI — Tier 1)
+- Problem: Rawhide CI was red on every push. 4 of the 6 failures were stale tests asserting behavior that production code intentionally changed, not real regressions.
+- Root causes:
+  - `avatar_repository_test.dart`: asserted DB `schemaVersion == 29`; schema is now 32 (bumped 3× without updating the test).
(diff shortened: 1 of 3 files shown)


MisterLotto:
I'm closing this and the partner PR. The hope had been that the two together could address lingering CI failures with the Rawhide branch, but more issues keep popping up unreliably which has me thinking there's more broken under the hood. There's a test teardown race where a disposed chat service still gets notified. Going to take a look at a more comprehensive fix for rawhide.

PR #46, opened 2026-06-07
fix(realism): correct card-seeded Bond doubling + realign stale one-shot test

MisterLotto:
Fixes 2 of the failing tests in `test/services/chat_service_realism_engine_test.dart`. Each was classified before fixing (stale test vs. real regression).

**1. Card-seeded Bond was doubled on every new chat (real bug — production fix)**
`startNewChat` seeded a character's initial Bond by running the card value through the legacy ±150→±300 score migration, which multiplies values of magnitude ≤150 by two. Card seeds are already authored on the ±300 scale, so a card set to a starting Bond of 55 began every new chat at 110. (`trustLevel` in the same block was never migrated, confirming the inconsistency; the migration is also nonlinear — 200 stays 200, 55 becomes 110.) The migration is only valid for previously-persisted *session* scores, where it's still applied. The seed now just clamps to range. Scoped to the 1:1 path; group seeding untouched, so Realism/Needs parity is preserved.

**2. One-shot eval prompt-marker assertion was stale (test-only)**
The test checked the one-shot prompt for markers from the wrong eval path. Verified the one-shot path fires and applies deltas correctly; updated the assertion to markers actually in that prompt.

**CI:** This branch fixes the 2 Realism Engine tests. The other 4 failures (avatar schema assertion + the `chat_message` swipeIndex cases) are handled by the separate stale-test fix branch. I verified on a local merge of both that all 6 originally-failing tests pass together, so once both land, the previously-red CI tests are green. (Independent diffs, disjoint test files — no interaction.)


MisterLotto:
I'm closing this and the partner PR. The hope had been that the two together could address lingering CI failures with the Rawhide branch, but more issues keep popping up unreliably which has me thinking there's more broken under the hood. There's a test teardown race where a disposed chat service still gets notified. Going to take a look at a more comprehensive fix for rawhide.

PR #47, opened 2026-06-08
fix: green Rawhide CI — bond-doubling regression, dispose crash, + stale/flaky tests

MisterLotto:
**Why this is one PR instead of separate small ones:** the failures were interconnected in a way that defeated the usual one-fix-at-a-time approach. Two earlier targeted PRs were opened and closed because neither greened CI on its own, and fixing the deterministic failures *uncovered* intermittent flakes that had been masked underneath them — a test-teardown race (`notifyListeners` on a disposed service) and a Chaos-Mode wheel that hangs headless tests. Splitting these back apart would mean each piece shows red CI until the others land, and the flakes only reproduce reliably when everything runs together. So this is deliberately one coherent pass to take Rawhide CI from red to green in a single, verifiable step. Full suite passes locally (890) and on CI.

**Production fixes**
- **Card-seeded Bond no longer doubled on new chats.** `startNewChat` ran the legacy ±150→±300 score migration on the card's initial bond, but card seeds are already on the ±300 scale — so a card authored at 55 started every new chat at 110. Trust (never migrated) confirmed the inconsistency. Seed now just clamps; migration stays on the legacy session-load paths. 1:1-only, group seeding untouched.
- **No more "used after disposed" crash when closing a chat mid-load.** `ChatService` async work (e.g. objective loading) could call `notifyListeners()` after disposal. Added a `_disposed` guard via central override — pre-dispose behavior unchanged.

**Test fixes**
- Schema assertion 29→32; three `chat_message` cases realigned to the `ChatMessage` swipeIndex clamp-to-0.
- One-shot eval test asserted markers from the wrong prompt — realigned (the path itself works).
- Disabled Chaos Mode in a group test that didn't assert it: when chaos auto-fired, `sendMessage` awaited a wheel completer only the UI resolves, hanging headless tests ~30s intermittently. Not a prod bug.


linux4life1:
**Ported to current Rawhide (post god-file modularization / PR #48)**

The two production fixes from this PR (#47 by @MisterLotto) have been integrated into Rawhide:

- **Card-seeded Bond no longer doubled on new 1:1 chats.** V2.5 card `FrontPorchExtensions` short/longTermBond values are authored on the current ±300 scale. The legacy ±150→±300 migration was being erroneously applied on every fresh card-seed path (`setActiveCharacter` 0-session + `startNewChat` 1:1 ext branch) via the extracted `RelationshipService.seedFromV2OrExt`. Port uses a dedicated `seedFromCardV2OrExt` (plain clamp only). Legacy persisted session data still auto-migrates when appropriate. Group seeding untouched. 1:1/group parity preserved exactly.

- **No more "used after disposed" crash when closing a chat mid-load.** Added central `_disposed` guard + overrides on `notifyListeners` and `dispose` in `ChatService` (protects all async DB-then-notify paths like objective loading without per-site checks). Pre-dispose behavior unchanged.

Changes were adapted to the post-Stage-3 extracted leaf architecture (relationship_service owns the seed helpers/migration surface; god only thin calls + the guard) while preserving *all* behavior, reset hygiene, "incomplete zeroing of secondary config on group/0-session/new-chat now complete", both startNew branches, full "keep reset blocks in sync" comment lists (current leaves + cross-refs), dedicated test coverage only, aug/integration tests untouched (per the "aug exercising only passive/qualified (no leaf-specific aug file edits...)" rule), 0 new god private _ methods (void _ count stayed exactly 15 live after every edit + final), deletion of dead code as part of the task, claims vs on-disk exact, and every other god-file/CLAUDE.md/AGENTS.md rule.

**Full credit to the original author:**
Commit authored as MisterLotto (exact name+email from the original PR #47 head 8f67d793aa7c8431dcc42753813181dc7958dfa2) + explicit Co-authored-by + multiple references in the commit message body and `.claude/changelog.md` entry. This gives @MisterLotto primary authorship credit in git, GitHub UI, blame, and contributor stats while transparently documenting the port/adaptation.

See:
- Commit: fe91830b4ef2c99fc738e3e79c038a0cb03051f0
- `.claude/changelog.md` top entry (detailed port notes, verification raws, hygiene summary)
- `docs/Rawhide.md` (user-facing bullet)

This supersedes the original changes on the pre-refactor tree. Thanks for the original CI-stabilizing work, @MisterLotto!

Closing as the fixes are now in mainline Rawhide.

PR #50, opened 2026-06-12
feat(group): fork-to-group wizard with per-character custom entrances

MisterLotto:
## Description
Replaces the single "Fork to Group Chat" popup with a stepped wizard (matching the existing create-group wizard pattern), and adds an **optional per-character entrance** for each character you bring in when forking a 1:1 chat into a group.

Each added character can be given one of two entrance modes, or none:
- **Opening line** — your text is used as that character's entrance **exactly as written** (no LLM generation).
- **Direction** — your text is a hidden note; the **AI writes** the entrance in the character's own voice (the note never appears as a message).
- **Blank** — the character is silently inserted into the turn order.

Turn order after the arrivals: the group's rotation becomes `[original participant(s), arrivals with an entrance (in add order), arrivals without an entrance]`. Entrants take their entrance turns in order, then the conversation resumes with whoever falls **right after the last entrant** in that rotation (the first silent arrival, or wrapping back to the original if there are none). Random turn order is unaffected.

Forking is **1:1 → group only**: the menu item is hidden inside group chats and `forkToGroupChat` guards against running from a group (it would otherwise rebuild a group from just the active speaker). Adding members to an existing group remains the job of **Add Character to Group**.

**Note:** This PR targets `Rawhide` (the active development branch for all new features, per CLAUDE.md / AGENTS.md). Desktop only — the matching web-UI changes are a follow-up; bundling them here was too much for one reviewable change, so the web fork is unchanged (no regression).

## Type of Change
- [x] New feature

## Testing
- [x] Unit tests added/updated (`group_entrance_test.dart`, 8 tests: forced entrance over round-robin/random, single + multi-entrant cut-in turn order, silent join, random no-op, verbatim Opening-line contract)
- [x] `flutter analyze` clean
- [x] Manual testing completed (Opening line verbatim, Direction AI-authored, multi-character next-turn order, fork-from-group blocked)

## Checklist
- [x] Code follows style guidelines (matches surrounding style)
- [x] Linting passes (flutter analyze)
- [x] Tests pass (new tests green)
- [x] Documentation updated (user-guide.md, Rawhide.md)
- [x] No breaking changes


linux4life1 (changes requested):
Hey @MisterLotto — thanks for this PR! The wizard design is solid, the file decomposition is clean, the tests are well-focused, and the documentation is thorough. This is clearly a well-thought-out contribution and a real UX improvement over the old dialog. A few things need to be sorted before we can merge.

---

## Required Changes

### 1. Remove `build_and_deploy.ps1`

This file has hardcoded paths specific to your local machine (`H:\Flutter\bin`, `H:\Front Porch AI`) and isn't related to the fork-to-group feature. Per project guidelines, scripts like this belong in `scripts/` and need to be discussed before landing. Please drop it from this PR — if you'd like to contribute a portable deploy script, a separate PR would be the right place for it.

### 2. `_entranceDirective` mutable field — race condition risk

```dart
String? _entranceDirective;
```

This field is set on `ChatService` (a singleton) inside an `unawaited` async block, then consumed inside `_buildPrompt`. Because the entrance loop runs fire-and-forget, there's a window where a user-triggered generation could consume or overwrite the directive before the correct entrance generation reaches it.

At a minimum, please guard the entrance block so no user-generated turn can start while entrances are in flight — the simplest approach would be checking/setting `_isGenerating`. The cleaner long-term fix would be passing the directive as a local parameter to `_generateResponse` rather than storing it on the instance, but we'd understand if you want to take the simpler guard path for now and note the follow-up.

### 3. Rebase onto current Rawhide

Rawhide has moved forward since you submitted this — there are now merge conflicts in `chat_service.dart`, `chat_page.dart`, `Rawhide.md`, and `changelog.md`. Since there are other required changes to address anyway, could you rebase onto current Rawhide when you push the fixes? That keeps the diff clean for final review.

---

## Should Fix (before or shortly after merge)

### 4. Entrance map keyed by character name

```dart
Map<String, ({String text, bool creative})> entrances  // keyed by name
```

If two characters in the library share a name, the second one's entrance will silently overwrite the first and the `firstWhere` lookup becomes ambiguous. You're already using `stableGroupId` elsewhere in this same PR for identity — could you switch the map key to that?

### 5. Entrance generation failures are swallowed silently

```dart
} catch (e) {
  debugPrint('[Fork:Entrance] generation failed: $e');
}
```

If a Direction-mode entrance fails (LLM timeout, backend error), the user just sees the group load with no indication that anything went wrong. A `SnackBar` or in-chat system message along the lines of `"⚠ {CharName}'s entrance could not be generated"` would give the user something to act on.

### 6. Entrance-step hint text is only correct in one case

```dart
'After all entrances, ${originalName ?? "the original character"} speaks next...'
```

This is only true when *every* arrival has an entrance. The actual rule (which the code implements correctly!) is that the next turn goes to whoever falls after the **last entrant** in the rotation — which is the first *silent* arrival if there is one. The hint text would be more accurate as something like `"After all entrances, the next turn follows whoever falls right after the last arrival in the rotation."` — or just omit the hint for the "all entrances" case and show it only when there are silent arrivals.

---

## Nits (no blocker)

- **Prompt injection**: `$text` from the user is interpolated directly into `_entranceDirective` without sanitization. Low risk since it's the user's own chat, but worth noting for prompt hygiene if you have an easy fix.
- **Image path resolution**: `forkAvatar` uses `FileImage(File(c.imagePath!))` directly; the old dialog used `_resolveCharImage()`. If `imagePath` can ever be relative, this could break — worth a quick sanity check.
- **`RadioGroup<bool>` Flutter version**: Just double-check this compiles on your local Flutter SDK. It's a relatively recent Material 3 addition — if you hit a minimum SDK issue, the project already uses `RadioListTile` with manual `groupValue` as a fallback pattern.
- **`const` opportunities**: A handful of `Text` widgets with string literals inside the new step files could be `const`.

---

Thanks again for the contribution — the wizard itself is great and we're looking forward to landing it. Happy to discuss any of the above if something isn't clear.

MisterLotto:
Yeah, sorry that build and deploy is my local tool. Wasn't supposed to get included. I'll work on the other stuff.

linux4life1:
All good, thank you for everything thus far. Don't know if your on the discord server or not, but if not after your PR I am going to cut the branch to 0.9.9-Beta so stabilization can get ironed out and then pushed to main. I think we have a good base thus far and I really want to get your feat into the new stable!

linux4life1 (approved):
All requested changes addressed — build script removed, race condition guarded with `_entrancesInFlight`, entrances keyed by `stableGroupId`, failures surfaced in-chat, hint text corrected, image path resolution fixed, and cleanly rebased with zero conflicts. 

Great work @MisterLotto — looking forward to getting this into 0.9.9-Beta! 🎭

PR #53, opened 2026-06-14
Windows-specific UI ghosting issue fix

S-A-M-F:
As per f975e66d7213d28af181a966732fcdbb04c4c9ef description:

Root cause: on Windows, window_manager's Maximize() posts WM_SYSCOMMAND
SC_MAXIMIZE (async) while SetBounds/SetWindowPos is synchronous.
When restoring maximized state, the old code called setSize() with
the maximized screen-filling dimensions, briefly putting the window
at full-screen in non-maximized state. A frame rendered in the async
gap before maximize processed, and the transparent acrylic background
(ACCENT_ENABLE_TRANSPARENTGRADIENT) let DWM composite the stale buffer
instead of discarding it — producing the ghost.

Fix:
- Track non-maximized window bounds via onWindowResized/onWindowMoved
  and an initial capture in the first-frame callback.
- On close while maximized: save the tracked normal bounds, not the
  current maximized (screen-filling) size.
- On restore while maximized: setBounds(normalRect) then maximize(),
  so the window never transitions through a full-screen non-maximized
  intermediate state.
  
Thanks to [MisterLotto](https://github.com/MisterLotto) for initial discovering of the issue's origins and conclusive testing of fix.


linux4life1 (changes requested):
Thanks for the PR, S-A-M-F! The description was really clear on the root cause — the async WM_SYSCOMMAND SC_MAXIMIZE + setting a full-screen rect before the maximize took effect, combined with the acrylic background and DWM composition on Windows. Appreciate you hunting this down and shipping a fix along with the required docs/Rawhide.md update.

The overall direction is solid: track the actual non-maximized ("restore down") bounds separately, and on a maximized restore do setBounds with the normal rect first, then maximize(). That prevents the intermediate full-screen non-max state that was causing the ghost.

The changes handle the reported scenario well. Before landing, there are a few real bugs and robustness issues I'd like to see addressed:

1. The new `_normalWidth`, `_normalHeight`, `_normalX`, and `_normalY` fields are only ever populated from live windowManager queries (the post-frame capture when not maximized, or the onWindowResized/onWindowMoved listeners). They are never seeded from the persisted prefs values on startup.

   This is a problem for the common "launch while maximized" case. The post-frame capture explicitly skips when `isMaximized()` is true after the maximize call. So if the user then closes while still maximized, the onWindowClose max branch will write the hardcoded defaults (1280x720 + 0/0) instead of the user's real previous normal size and position. Their window geometry gets lost.

   **Fix needed:** In the post-frame callback (early, before or instead of the conditional live capture), unconditionally seed the `_normal*` fields from the four stored prefs values if they are all non-null. The live listeners can still update them when the user actually resizes or moves while in the normal (non-max) state. This way a max-launched session still has the correct numbers in memory for a same-session close-while-max.

2. The three new capture sites — the post-frame block, `onWindowResized`, and `onWindowMoved` — call `windowManager.isMaximized()`, `getSize()`, and `getPosition()` with no surrounding try/catch.

   The original restore logic in `waitUntilReadyToShow` and the entire onWindowClose handler both wrap their windowManager + prefs work and do debugPrint on error. An early transient failure (platform not ready, DWM hiccup, etc.) in the new paths would leave the tracking fields at defaults and could silently corrupt the persisted normal rect on the next close.

   **Fix needed:** Wrap the capture sequences in try/catch + debugPrint in all three places, matching the pattern used in the rest of the window state code.

3. Users upgrading from a pre-PR build who previously closed while maximized will have the old full-screen rect stored in the `window_*` prefs (with `window_maximized=true`).

   On next launch the maximized restore path now requires all four values and will either skip the setBounds or feed a fullscreen rect into it before calling maximize(). This can reproduce the exact ghosting symptom on the first relaunch after upgrading, until the user un-maximizes once and lets the listeners capture a real normal rect.

   **Fix needed:** Add a defensive guard in the maximized restore path. If the loaded width/height look suspiciously large (or match typical maximized dimensions), just call `maximize()` without the preceding setBounds, or fall back to letting the OS supply a reasonable restore rect. This protects the upgrade experience.

A couple of smaller items:

- In the max branch of onWindowClose: `await prefs.setDouble(_k('window_x'), _normalX ?? 0);` (and same for Y). The fields are `double` and the setter expects doubles — `?? 0.0` would be slightly cleaner and more consistent.

- The listeners only cover resized and moved. The WindowListener interface also has onWindowUnmaximize (and onWindowMaximize). In practice the resized/moved events after an un-max should fire while not maximized, but if testing on Windows shows cases where the normal rect isn't captured after title-bar un-max, Win+Down, etc., adding an explicit `onWindowUnmaximize` that does the same capture would be more robust. This one is optional but worth a quick test.

- The docs/Rawhide.md bullet is accurate, in the right place, and follows the per-branch policy. Very minor optional polish: the parenthetical explanation is a little implementation-detail heavy compared to some of the other bullets. Something shorter like "— no more ghost frames when reopening after closing while maximized" would be friendlier for users skimming the update notes.

Overall this is a good, focused fix for a real Windows-only annoyance. The tracking approach is the correct long-term solution. The main things above are about making the normal-bounds state reliable across "launch maximized", "close while maximized", and "first run after upgrading from a version that saved the maximized rect".

Let me know what you think or if any of these don't match what you saw in testing. Happy to discuss.

Thanks again for the contribution!

S-A-M-F:
Ok, read it through. 
For issue 1: I can do that, but:
- as much as I tried - I was actually unable to reproduce window geometry loss on test build in practice - probably due to another windows quirk - even while I see your logic, and somehow perplexed by the fact it works where it should not. I guess on MacOS the bug is user-observable?
- solving it makes issue 3 way more severe than just one-time upgrade issue - that WAS intentionally left as-is exactly due to that fact. But I would like to have your confirmation that it's actually observable issue on non-windows - even if I agree that code looks wrong in theory, and thus probably should be fixed. 

For issue 2 - no objections here, that's proper.

For issue 3:
- it was intentionally left as-is, for simplicity. Possible way, considering the issue 1 fix, will be to add another parameter, tracking the version of the code that wrote window size/position - then leave it as is here as artifact of the past (which is unclean as it would serve no real purpose beyond the smoother one-time migration that is self-healing - I hardly can imagine another use case for this - so YAGNI kinda applies) or add it temporarily (and still get same issue with anyone who skips the version with upgrade mechanism present). That was considered overengineering and thus omitted - but it certainly worth it if it's actually needed - as fixing issue 1 would require that. But exact way to do it... I'm not sure, for abovementioned reasons.
- Workaround with "typical maximized dimensions" will be even more of an overengineering - and fragile - because maximized window size can depend on size of taskbar, any other custom sidebars, user having taskbar on side of the screen, or not having it at all on this screen on multi-monitor configs, having monitor in album orientation, using a custom tiling WM that adds spacing around even maximized window, any combination of the above, etc. 

For minor issues I will take care of that separately. 

So, I will do actual changes in a few hours or so - this day is unexpectedly filled. Thanks for pointing that all out.

S-A-M-F:
On an afterthought, leaving theoretically incorrect code as is just because it works - indeed feels wrong. So, treat it more like an explanation why it was incorrect in first place, not why it should stay as it is. I'll think that through when I return to that later - need to make pause to let it all sink in.

linux4life1:
Those were just the issues Grok found with the PR code, the ghosting issue doesn't exist on MacOS so I can't debug/test. Since you run the affected OS (Windows) I defer to you on if we should merge as is or tweak the PR as suggested by the LLM. Breakage is expected in Rawhide so we can tweak it a bit as needed before cherry picking into main for next point release.

S-A-M-F:
I know. Ghosting issue is not present on my PC too (it's not only windows but also GPU driver-specific it seems), but [MisterLotto](https://github.com/MisterLotto) confirmed it was real on his PCs and that fix solved it. I wanted you to test other thing - namely - is that fix breaks window size persistence on MacOS as per issue 1? Because on Windows it is not - on my PC at least - even while looking at the code may suggest that it does.

Because if no immediate fix needed on 1 - then fixing 3 also is optional - since it's one-time self-healing issue. But 2 is a solid case. 

Thereupon, I shall make it a draft, just so it's clear I'm not fixed it yet.

linux4life1:
I just tested the PR branch on MacOS and there was no negative effects on my OS. will need to setup a Linux VM to make sure there are no ill effects on the penguin side

S-A-M-F:
Ok, for now addressed minor issues and issue 2 where applicable. Will wait for confirmation about linux - but if all is good - going to leave it as is with commentaries hinting at the issue and possible solution. On confirmation - making it ready to merge adain. Or probably just do it now - so you can merge whenever you find out it's ok.

linux4life1:
Didn't see any issues with Debian 12 VM in UTM. safe to merge

PR #60, opened 2026-06-20
feat(characters): export cards as JSON + fix raw .json import

MisterLotto:
## What & why

Characters could only be exported as PNG (with the V2 data embedded in the
`chara` chunk). There was no way to get a plain `.json` card out — even though
`.json` is the format people actually pass around (SillyTavern, Chub, etc.).

While digging in, I found a latent bug: the import file picker advertised
`.json`, but importing one never actually worked. `V2CardService.readCard`
only parses PNG bytes, so a raw `.json` fell through to the name-only fallback
and you got an empty character with just a filename.

This PR adds JSON export **and** makes JSON import real, so the two form a
genuine round-trip. No new format is invented — it's the standard
**Character Card V2** spec (`chara_card_v2`), the same JSON already embedded
in exported PNGs.

## Changes

- **`V2CardService`**
  - `saveCardAsJson` — writes the canonical envelope
    `{ "spec": "chara_card_v2", "spec_version": "2.0", "data": { ... } }`.
  - `readCardFromJsonFile` — imports a standalone `.json` card.
  - Extracted the inline parse logic from `readCard` into a shared
    `parseCardJson`, so the PNG and JSON paths decode identically instead of
    duplicating field mapping.
- **`CharacterRepository.importCharacter`** — detects `.json`, parses it as
  text, and persists without a source image so a placeholder avatar is
  synthesized (JSON carries no image). Also fixed the fallback filename to use
  `p.basenameWithoutExtension` instead of `split('/')`, which was wrong on
  Windows paths.
- **UI** — "Export JSON" context-menu item on the character menu (next to the
  existing "Export PNG"), wired to a new `_exportCharacterJson` handler
  mirroring the PNG one.

## Notes

- The new JSON logic lives in `v2_card_service.dart` (well under the 500-LOC
  cap). The larger files only gained thin menu/handler wiring next to their
  existing siblings.
- Lorebook entries export with both `keys[]` (array) and `key` (comma string),
  so cards round-trip into SillyTavern/Chub *and* back into Front Porch.
- `extensions.front_porch` (Realism Engine settings) is preserved on
  round-trip; unknown third-party extension keys are also preserved per spec.

## Testing

- New "JSON Export/Import" test group in `v2_card_service_test.dart`: envelope
  shape, full no-avatar round-trip, flat legacy-V1 parse, malformed → null,
  and PNG/JSON parity. **18/18 pass.**
- `flutter analyze` clean on changed files.
- Manually verified in a Windows build: exported a character to `.json`,
  re-imported it, all fields + lorebook intact.


linux4life1:
Thanks @MisterLotto — this has landed on `Rawhide` 🎉

Your original commit was brought in **as-is** (merged, not rewritten) via #61, alongside two small review follow-ups:

1. **Unified the export envelope** — `saveCardAsJson` already wrote the proper V2 envelope (`{spec: chara_card_v2, spec_version, data}`), but PNG export still embedded the legacy *flat* data map with no `spec` marker, so the two formats weren't equivalent. Both now route through a shared `_buildCardV2Envelope` helper, so a Front Porch PNG carries the same standard marker as the `.json` export (better SillyTavern/Risu/Chub interop).
2. **Added a repository-level import test** — covers the user-facing half of your fix end-to-end: a raw `.json` routes through `readCardFromJsonFile` and gets a synthesized placeholder avatar instead of a name-only empty character.

CI (analyze + full test suite) is green on the merge. Closing this in favor of #61 since the code is now in `Rawhide` — full credit to you for the feature and the latent import-bug catch. 🙏

---

PR #62, opened 2026-06-22
Feat/kcpps generator

S-A-M-F:
Adds a "Generate kcpps" dialog that produces a KoboldCPP preset file with optimal settings for the selected model, backed by a thoroughly reworked VRAM estimator that now correctly handles mixed-attention MoE models (tested on Gemma 4, Qwen, LFM).

What's included

### MoE + mixed-attention VRAM estimation

- Parses per-layer head_count_kv arrays from GGUF metadata to distinguish full-attention layers from sliding-window attention layers
- Reads full_attention_interval (Qwen), sliding_window, swa_head_dim, MoE expert counts, and the full 16 MB header (incl. tokenizer vocab size)
- KV cache estimation now uses real per-layer head counts — verified against actual KoboldCPP output for Gemma 4 at 8K and 24K context (exact MiB match)
- SWA cell count caps at min(contextSize, 8×slidingWindow)/4 + sw/8 to prevent unbounded growth at large contexts
- GPU weight accounting includes always-on-GPU layers (embeddings + lm_head) when offloading experts
suggestBatchSize() helper finds the largest batch fitting within VRAM budget

### "Generate kcpps" dialog (lib/ui/dialogs/generate_kcpps_dialog.dart)

- Model selector, context size, KV quantization, batch size, threads, context mode (FastForwarding/SWA), greedy allocation toggle
- Real-time VRAM breakdown bar with 500 ms debounce on text fields
- Auto-suggest for batch size based on available VRAM
- GPU info tile showing adapter name, driver, dedicated/shared memory
- Wired into Settings page under Model Management 

### Refactored GGUF reader and widgets

- Extracted GGUFFileReader (shared binary reader) from the monolithic gguf_parser.dart
- Split GGUFModelInfo into its own file; VramEstimateBreakdown typedef into vram_estimator.dart
- Extracted reusable widgets: VramUsageSection, ContextManagementSelector, GpuInfoTile

### Documentation

- docs/moe-vram-estimation.md — deep-dive analysis of MoE VRAM over-estimation and KoboldCPP's autofit algorithm (319 lines)
- docs/Rawhide.md — user-facing "What's New" entry
- 35 new tests covering mixed KV heads, capped SWA, interval patterns, MoE offloading ratios (still not enough though...)


linux4life1:
## Review: kcpps generator + MoE VRAM estimator

Did a deep pass on the VRAM/GGUF logic and the dialog/widgets. The **core estimator is genuinely strong** — per-layer KV-head parsing, ISWA compression, and the MoE expert-offload GPU ratio are real improvements, and the tests pin concrete MiB values. The findings below are mostly in the UI layer plus a couple of backend correctness items. Verified each one against the branch.

### 🔴 Must-fix (correctness)

1. **Work after the dialog is popped** — `lib/ui/dialogs/generate_kcpps_dialog.dart:252-266`. On success we `Navigator.pop(true)` (252), then call `ScaffoldMessenger.of(context).showSnackBar(...)` (254) on the just-popped context, then the `finally` runs `setState(() => _generating = false)` (266) on a tearing-down widget. The snackbar silently no-ops and the `finally`/`catch` setStates are unguarded. → Capture the messenger **before** pop (or fire the snackbar from the parent), and guard the `finally`/`catch` with `if (mounted)`.

2. **`TextEditingController` created inside `build()`** (allocated every rebuild, never disposed, and resets/fights user input on any `setState` such as the 500 ms VRAM debounce):
   - CPU Threads — `generate_kcpps_dialog.dart:411` `controller: TextEditingController(text: '$_threads')`
   - SmartCache slots — `lib/ui/widgets/context_management_selector.dart:88` `controller: TextEditingController(text: '$smartCacheSlots')`
   → Promote both to `initState`-owned, `dispose`-d fields.

3. **Dead code: `_smartCacheSlotsController`** — `generate_kcpps_dialog.dart:37` (declared) + `:75` (disposed), but it's **never attached to any widget**; the selector makes its own inline controller (`:88`) instead. Delete the orphan and wire the owned controller into the selector — that also resolves the second half of #2.

4. **`setState` after `await` with no `mounted` guard** — `generate_kcpps_dialog.dart:116-126` `_refreshModelInfo`: `await mgr.getModelArchitectureInfo(...)` → `setState(...)`. Throws if the dialog closes mid-await. Add `if (!mounted) return;` after the await.

5. **AppColors violation in new UI** — the new "Generate KCPPS" button in `settings_page.dart` uses `Colors.white70` / `Colors.white.withValues(...)`. Per repo rules new UI must use `AppColors.textSecondary(context)` / `AppColors.borderOf(context)`.

6. **Dead color tier** — `lib/ui/widgets/vram_usage_section.dart:73-78`: `fraction > 0.85` and `fraction > 0.6` both assign `AppColors.logWarn`, so the high-usage tier is meaningless. Give `>0.85` a distinct (more alarming) color, or collapse the branches.

### 🟡 Please verify (likely bugs)

7. **`quantkv` is written as a string, KoboldCpp expects an int** — `lib/services/kcpps_generator_service.dart:100` puts `'quantkv': kvQuant` where `kvQuant` is `'f16'`/`'q8_0'`/`'q4_0'`. KoboldCpp's `--quantkv` takes an **integer index** (`0`=f16, `1`=q8_0, `2`=q4_0). If the kcpps loader doesn't coerce it, KV quant silently won't take effect. Map the string → `0/1/2` before writing. (FA-on via `noflashattention: false` is correct, since quantkv>0 needs flash attention. 👍)

8. **645-line dialog exceeds the 500-line cap** — the input form (model/context/threads/batch/kvQuant, ~317-471) is a clean extraction into a `KcppsBasicSettingsForm`.

9. **`gguf_reader.dart:99`** — `(data.getFloat32(...) as num).toInt()` is an upcast (`double` is already `num`); likely trips `unnecessary_cast` on the changed file. Drop the cast.

10. **`gguf_reader.dart:172` latent buffer-offset bug** — `_readScalar` case 8 uses `data.buffer.asUint8List(offset, strLen)`, where `offset` is view-relative, not buffer-relative. Safe today (every caller passes a fresh `raf.read()` buffer at offset 0) but inconsistent with the key decode at `:63` (`bytes.sublist`). Since `general.architecture` is read as a string scalar, this would silently corrupt parsing if anyone ever passes a sub-view. Prefer `bytes.sublist(offset, offset+strLen)`.

### 🖥️ Cross-platform / GPU backend coverage

`detectGpuBackend` (`kcpps_generator_service.dart:60`) is a good start but a few gaps worth a follow-up:

- **Vulkan only emitted on Windows.** On Linux with neither CUDA nor ROCm (Intel Arc, some AMD setups) it returns `{}` → CPU-only. KoboldCpp Vulkan works on Linux too — add a Vulkan fallback there. (`HardwareInfo` has no `hasVulkan` field yet; today the Windows path is a blind guess. Detecting Vulkan availability — `vulkaninfo`/`libvulkan` presence — would make this non-blind on both OSes.)
- **Device index is hardcoded to 0** — `usecublas: ['normal', 0]` and `usehipblas: [0]`. Multi-GPU users can't target a card or split. A "GPU device / tensor_split" selector (the GPU info tile already enumerates adapters) would be a solid value-add.
- **ROCm on Linux:** VRAM detection via sysfs `mem_info_vram_total` is nice. The common pain point is unsupported gfx versions needing `HSA_OVERRIDE_GFX_VERSION` — surfacing that hint (or a warning) for older AMD cards would save users a lot of grief. cuBLAS `'normal'` is a fine default; consider exposing `lowvram` when the fit status is "tight".
- **macOS/Metal:** `detectGpuBackend` correctly returns `{}` (Metal is automatic), but the VRAM budget on Apple Silicon comes from `ramMb * 0.75` (unified memory). Worth a short note in the breakdown that the "VRAM" bar is a unified-memory estimate on Metal, and that `usemmap`/`usemlock` behave differently there.
- **`wmic` is removed on recent Win11 builds** (`detectPhysicalCores`) — the `Platform.numberOfProcessors` fallback covers it, but core detection silently degrades to logical-core count. A PowerShell `Get-CimInstance Win32_Processor` fallback would keep SMT detection working.

### 🧪 Test coverage suggestions

- `vram_estimator_test.dart` only asserts `kvCacheMb`/`weightsMb` — **nothing covers `computeBufMb` or `overheadMb`**, which is the least-principled part of the estimate (the `nVocab*2` term is folded into per-batch-token cost and likely over-estimates at large batch). Add at least one `totalMb`/`computeBufMb` assertion.
- **No tests for `KcppsGeneratorService` at all.** The SWA vs FastForward branch in `buildKcppsContent` (the `noswa`/`nofastforward`/`noshift`/`smartcache` flag matrix) and `detectGpuBackend`'s backend mapping are pure functions — easy, high-value unit tests that would have caught #7.
- The mixed-head classifier ("smaller `n_kv × key_length` = full-attention layer", `vram_estimator.dart:324`) is an empirical heuristic validated only on Gemma 4 / Qwen / LFM. A short code comment flagging it as empirical + a test for a non-matching head ordering would harden it.

### Value-add / TODO ideas
- Persist last-used settings (context/kv/threads) per model.
- Show the resulting kcpps as a preview/diff before writing, and a "this overwrites X" confirm.
- "Tight fit" should suggest dropping KV quant or context, not just color the bar.

Overall: solid feature, mergeable after the 🔴 items + #7. The estimator math is the standout. Happy to push fixes for any of these if useful.

<sub>🤖 Automated review via Claude Code.</sub>

---

S-A-M-F:
- 1,2,3,4,6 - Yes, expected something like that. Good to have it pointed out.
- 5 - Umm... 71 case of Colors.white* variants alone in that file. Looks like a thing that should be done separately from that PR. As of now - I just added something that wasn't looking out of place here.
- 7 - nope. Koboldcpp accepts integers (and even floats that can round to them, apparently - though that was really unexpected to see - LOL, WHY?) as a legacy compatibility measure here - but it writes it's own configs with f16/bf16/q8_0/q5_1/q4_0 instead, other ways are actually aliased to that - checked it against koboldcpp sources. So it's ABSOLUTELY not an issue.
- 8 - Considered that before making PR - and not sure if I should. Extracted some genuinely useful elsewhere parts from that file already in fc2ab7c - but was stuck at this point - because furher extraction looked like increasing complexity for sake of arbitrary limit (15 callbacks doesn't look like a good thing either).
- 9 - Ok, right. Though no unnecessary_cast lint rule defined in analysis_options.yaml - hence linter not catched that. Maybe add it - so similar problems would be found not just here but in a whole project then?
- 10 - Yes, seems legit. Not an immediate concern, just as you noted - but futureproofing is due here.

> Vulkan only emitted on Windows.

Yes, this is as incomplete as it is. "Make it work, then make it right, then make it fast."

> Device index is hardcoded to 0

Multi-GPU users arguably don't need kcpps draft generator - they can do things on their own. Point was to provide a useful tool for someone who don't even know what config file is internally. Or create a sane draft without manually launching koboldcpp first. Good idea, but I guess it's not absolutely necessary for this to be useful for many already. 

> ROCm on Linux
> HSA_OVERRIDE_GFX_VERSION

Hmm... Should look into that. Really good point.

> lowvram

Not sure if it's practical. MoE models already use GPU bus noticeably while working - not loading KV cache into VRAM will likely further clog it... Need to do some research before deciding if that's really an added value. 

> macOS/Metal
> Worth a short note in the breakdown that the "VRAM" bar is a unified-memory estimate on Metal, and that usemmap/usemlock behave differently there.

Just copied that from existing auto-configurer. But whole thing becomes kinda pointless here - as MoE models do not provide any advantage on Mac - because there's no way to split them... Or maybe I'm just don't understand how that works. Need a good source of info about it before figuring out what to even put into that note. 

Alternatively - would be good if you did that part right. Would likely be more effective, as I honestly may lack specific knowledge to compensate for the inability to iterate on a real system to get the estimator near-perfect.

> wmic is removed on recent Win11 builds

Huck! Good heads-up. Aboslutely should do - even if the issue is at most one unutilised core. Probably should also consider if it necessary to distinguish E/P-cores on some intel CPU too.

> Tests

Agree with all 3 points. 

Anyway, that's a draft - and a fresh glance is exactly what I needed here. Will work upon that a bit later, I guess rebasing that won't be a big deal even if it comes to this - most changes are backend- or settings-related, and you seem to work on other part of app lately. Though if you fix any of what I confirmed as real issue before I do - that would be good too.

linux4life1:
Fixes for the confirmed items are ready as a single commit on top of your branch. The sandbox I'm running in can only reach `linux4life1/front-porch-AI`, not your fork, so I pushed it there:

**Branch:** `claude/pr62-review-fixes` (= your `fc2ab7c` + one fix commit `cefcf52`, fast-forward, no rewrite of your work)

To pull it onto this PR:

```bash
git fetch https://github.com/linux4life1/front-porch-AI.git claude/pr62-review-fixes
git cherry-pick cefcf52
```

### What `cefcf52` changes
- **`_generate` post-pop work** — capture messenger + navigator before the awaits, show the snackbar via the captured messenger (survives the pop), then pop; reset `_generating` only in `catch` under an `if (!mounted) return`; dropped the `finally`.
- **Inline controllers** — added a dialog-owned, disposed `_threadsController` (synced in `_initDetection`); `ContextManagementSelector` now takes the dialog's already-owned `_smartCacheSlotsController` instead of building its own (this also un-orphans that controller).
- **`_refreshModelInfo`** — `if (!mounted) return` after the `await`.
- **AppColors** — the new "Generate KCPPS Config..." button only (`textSecondary`/`borderOf`); your point about the file's ~71 pre-existing `Colors.white*` stands, left for a separate pass.
- **`vram_usage_section`** — `>0.85` tier now `logError` (was a dead duplicate of the `>0.6` warn tier).
- **`gguf_reader`** — dropped the `as num` cast; fixed the latent `_readScalar` string buffer-offset (now `data.offsetInBytes + offset`).

#7 (quantkv string) withdrawn per your source check 👍. #8 (file length) left as-is per your call.

Note: no Flutter SDK in my environment, so I couldn't run `flutter analyze` locally — verified by read + grep; your CI is the real gate.

---

### Follow-up: the macOS/Metal note you asked me to take

Researched the MoE-on-Mac question. Short version: the two ideas are separable —
- **MoE *architecture*** is a genuine win on Apple Silicon (decode is bandwidth-bound; only active experts are read per token — e.g. Qwen3-30B-A3B ~80–100 tok/s on M4 Max). Apple even benchmarks Qwen3-30B-A3B / gpt-oss-20b on their silicon.
- **Expert-offload-to-CPU** (`--n-cpu-moe` / `--moecpu`) exists to dodge a *discrete GPU's PCIe VRAM↔RAM split*. Unified memory has no such split, so it frees no memory on Mac and usually *slows* generation. Its only Mac use is squeezing under Apple's ~70–75% Metal GPU-allocation cap when a model otherwise won't load.
- **Footprint = total params** (all experts resident); active count only drives speed.

Implication for the estimator: on Metal, don't model expert-offload as VRAM savings (`gpuWeightRatioWhenOffloadingExperts` / `moeExpertsOnCpu` should no-op on macOS); budget the full quantized model against ~70% of total unified memory, and label the bar as a unified-memory estimate. Suggested user note:

> **macOS (Apple Silicon):** GPU and CPU share one unified-memory pool, so the entire quantized model loads together — offloading MoE experts to CPU frees no memory here and usually slows generation. This bar estimates unified memory; budget against ~70% of total RAM. MoE models still run fast on Mac because only the active experts are read per token.

Sources: llama.cpp [#15077](https://github.com/ggml-org/llama.cpp/pull/15077), [#11397](https://github.com/ggml-org/llama.cpp/pull/11397), [discussion #15396](https://github.com/ggml-org/llama.cpp/discussions/15396); Apple [WWDC20 unified memory](https://developer.apple.com/videos/play/wwdc2020/10686/). Happy to fold the Metal no-op + note into a follow-up commit if you want it.

<sub>🤖 Automated via Claude Code.</sub>

---

linux4life1:
**Update — Metal follow-up is on the branch too.**

Added a second commit `0af73d2` ("no-op MoE expert-offload on macOS + unified-memory note"). `claude/pr62-review-fixes` is now `cefcf52` + `0af73d2`, still a clean fast-forward of `feat/kcpps-generator`.

- Estimator call sites now pass `moeExpertsOnCpu: !Platform.isMacOS` (was hardcoded `true`). On Apple Silicon CPU+GPU share unified memory, so the whole quantized model is modelled as resident there instead of pretending experts are offloaded to CPU (which frees nothing on a Mac and would only slow generation).
- Done at the **call site**, not inside `VramEstimator.estimateFromArchitecture` — that function is pure and has a test passing `moeExpertsOnCpu: true` + asserting `weightsMb`, so baking `Platform.isMacOS` into it would pass on Linux CI but fail on macOS CI. Keeping the platform decision in the UI keeps the estimator deterministic.
- Added the macOS unified-memory note under the VRAM breakdown (budget ~70% of total RAM; experts not offloaded for MoE).

Pull both:
```bash
git fetch https://github.com/linux4life1/front-porch-AI.git claude/pr62-review-fixes
git cherry-pick cefcf52 0af73d2
```

<sub>🤖 Automated via Claude Code.</sub>

---

S-A-M-F:
Will merge that back to my branch and do analyse. Thank you.

S-A-M-F:
Only one preexisting warning not related to any code touched here with cefcf52 0af73d2 added on top.
warning - The value of the field '_userMessagesSinceLastEvolution' isn't used - lib\services\chat_service.dart:8005:7 - unused_field

linux4life1:
Merged into `Rawhide` as `ad8719c` (no-ff merge preserving your `56ac450`). Thanks @S-A-M-F — the estimator work is excellent.

Two files conflicted; here's how they were resolved:

- **`.claude/changelog.md`** — union of both sides (kept the Rawhide PR #66 / macOS-CI entries *and* your review-fix entry). No content dropped.
- **`lib/utils/gguf_parser.dart`** — the only real code conflict. While this PR was open, Rawhide landed `82dccf5`, which wrapped the **inline** metadata value-switch in `getModelArchitectureInfo` in a `try/catch → break` to stop a RangeError on large-vocab headers (`tokenizer.ggml.tokens` overrunning the header read). Your PR deleted that inline switch entirely and moved parsing into `GGUFFileReader.parseMetadataBytes` (4 MB → 16 MB read). Since Rawhide's only change to that file lived inside the hunk you removed, the resolution takes your version verbatim (verified byte-identical to the PR head for that file).

Rawhide's RangeError fix is **preserved by construction**: `parseMetadataBytes` already bounds-checks the `tokenizer.ggml.tokens` walk and returns the partial `meta` (the arch keys precede the tokenizer data, so parsing *succeeds* rather than throwing); `getModelArchitectureInfo` keeps its outer `try/catch` backstop; and the other half of `82dccf5` (the negative-parse cache in `model_manager.dart`) is in a file this PR doesn't touch, so it survived untouched.

One small hardening on top: the per-layer `head_count_kv` array loop in `parseMetadataBytes` was the one read lacking a per-iteration bounds check (every other read there has one) — added the matching `if (offset + elemSize > bytes.length) break;` guard so the merged reader is no less robust than Rawhide. For real models that array is tiny and well inside the 16 MB buffer, so it's defensive parity, not a behavior change.

Verification on the merged tree: `flutter analyze --no-fatal-warnings --no-fatal-infos` → **No issues found!**; `dart fix --dry-run` → **Nothing to fix!**; `vram_estimator` tests **35/35 green**. (The `_userMessagesSinceLastEvolution` warning you saw is gone — PR #66 reworked that file.)

Deferred by mutual agreement, not blocking: #5 (the file's pre-existing `Colors.white*`), #8 (the 645-line dialog vs the 500-line cap), and the GPU-detection / test-coverage follow-ups. Worth a future pass.

<sub>🤖 Merge + conflict resolution via Claude Code.</sub>

PR #67, opened 2026-06-24
fix(hardware): robust GPU detection for RTX 50-series (Blackwell) on Windows

bigsombar:
## Description

On a fresh Windows install with an NVIDIA RTX 5060 Ti 16GB, the **Settings → Hardware & GPU** panel shows `Unknown GPU` and the `Use CuBLAS (Nvidia)` chip is greyed out, so the user cannot enable CUDA acceleration. KoboldCPP then launches without `--usecublas` and falls back to CPU inference.

The previous Blackwell fix in `fc35d6a` made `nvidia-smi` the authoritative source for NVIDIA cards but still invoked it via a bare `Process.run('nvidia-smi', ...)`. On Windows the NVIDIA driver sometimes installs `nvidia-smi.exe` only under `C:\Program Files\NVIDIA Corporation\NVSMI\` **without adding that directory to PATH** — the call throws `ProcessException`, the `catch(_)` silently swallows it, and the user is left with `Unknown GPU` even though their card and driver are perfectly functional.

Several smaller issues compounded the failure:

1. **Registry check (Method 1) dropped `DriverDesc` when `qwMemorySize` was missing** — but early 596.x drivers for new architectures often populate the name without the 64-bit VRAM size, so we lost the GPU name too.
2. **WMI check (Method 2) picked the GPU with the highest `AdapterRAM`**, but `AdapterRAM` is `uint32` and overflows to 0 for any card with VRAM that is an exact multiple of 4 GB (8/16/24 GB). On Intel-iGPU + NVIDIA-dGPU systems the iGPU won the race and the discrete GPU was discarded.
3. **`nvidia-smi` CSV parsing** used `int.tryParse` on the whole VRAM column, so older `nvidia-smi` versions that ignore `--format=nounits` and emit a `MiB` suffix would parse as null and lose VRAM.
4. **Vendor detection** only matched `nvidia`/`geforce`/`amd`/`radeon`/`intel`/`iris`/`uhd` — cards marketed as `RTX 5060 Ti` without the `NVIDIA` prefix, or Intel `Arc`, would not be routed to the right backend.
5. **PowerShell `ConvertTo-Json`** can emit very large `Int64` values (e.g. `17_179_869_184` bytes for 16 GiB VRAM) in scientific notation on Windows PowerShell 5.1, which the previous `int.tryParse` didn't handle.
6. **Several `as int` casts** on WMI `AdapterRAM` / `SharedSystemMemory` would throw `TypeError` if `ConvertTo-Json` returned them as strings.

### Changes (all confined to `lib/services/hardware_service.dart`)

**New private helpers:**

| Helper | Purpose |
|---|---|
| `_runNvidiaSmi(List<String> args)` | Tries `nvidia-smi` on PATH first; on Windows falls back to `C:\Windows\System32\nvidia-smi.exe` and `C:\Program Files\NVIDIA Corporation\NVSMI\nvidia-smi.exe`. Returns `null` if the binary can't be found or returns non-zero. |
| `_parseNvidiaSmi(String stdout)` | Splits CSV on the **first** comma (handles workstation cards with commas in the name), strips any trailing unit suffix via `^(\d+)` regex, picks the entry with the largest VRAM. |
| `_vendorFromName(String name)` | Expanded patterns: `rtx` / `gtx` / `quadro` / `tesla` (Nvidia), `radeon` / `firepro` / `instinct` (AMD), `arc` (Intel). |
| `_NvidiaSmiResult` (private class) | Tiny value carrier for the parser; kept in the same file because it's tightly coupled to `_parseNvidiaSmi`. |

**`_detectWindows()` restructure:**

- **Method 0 (NEW)** — try `nvidia-smi` FIRST. For NVIDIA cards this is now authoritative; we never need to fall through to the broken registry/WMI path on a working driver install.
- **Method 1 (Registry)** — track a separate "named candidate" so we still recover `DriverDesc` when no item has a usable `qwMemorySize`. Handle `int`, `double`, and `String` for `qwMemorySize` (the latter two via `double.tryParse` for scientific notation).
- **Method 2 (WMI)** — prefer discrete GPU vendors (NVIDIA/AMD) by name before falling back to max-`AdapterRAM` selection. Coerce `AdapterRAM` to `int` safely (no more `as int` casts).
- **Method 3 (nvidia-smi final sweep)** — run whenever VRAM, name, OR vendor is still unknown, so we fill in any remaining blanks.

**`_detectLinux()` improvements:**

- When `lspci` reports both a VGA controller (typically the Intel iGPU on hybrid laptops) and a `3D controller` (the discrete NVIDIA/AMD GPU), prefer the `3D controller` — the iGPU is irrelevant for LLM inference.
- Use `nvidia-smi` for the GPU **name** as well as VRAM when vendor is NVIDIA.

**`_checkDrivers()`:**

- Use `_runNvidiaSmi([])` so CUDA is detected even when `nvidia-smi` is not in PATH.

### Verification on RTX 5060 Ti 16GB (Windows)

I tested this on my machine: RTX 5060 Ti 16GB, Windows 11, driver 596.x. Before the patch the panel showed `Unknown GPU` and the CuBLAS chip was disabled. After the patch:

- Settings → Hardware & GPU shows `NVIDIA GeForce RTX 5060 Ti`
- VRAM gauge shows `16384 MB`
- `Use CuBLAS (Nvidia)` chip is enabled and auto-selected
- KoboldCPP launch args include `--usecublas 0`
- Console shows `[Hardware] nvidia-smi (Method 0): NVIDIA GeForce RTX 5060 Ti, 16384MB`

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring

## Testing

- [ ] Unit tests added/updated
- [x] Integration tests pass
- [x] Manual testing completed

### Manual testing performed

- Built and ran on Windows 11 with an RTX 5060 Ti 16GB (driver 596.x).
- Verified Settings → Hardware & GPU now shows the correct GPU name and VRAM.
- Verified `Use CuBLAS (Nvidia)` chip is enabled and auto-selected.
- Verified KoboldCPP launches with `--usecublas 0` (visible in the Kobold log dialog).
- Verified generation runs at CUDA-accelerated speeds (vs CPU fallback before the patch).

### Notes on unit tests

I did not add new unit tests because `HardwareService` spawns real OS processes (`nvidia-smi`, `lspci`, PowerShell) and is not currently mockable — there is no `ProcessRunner` abstraction to inject. The existing `FakeHardwareService` in `test/golden/support/fakes_services.dart` continues to work unchanged because it overrides only the public interface, which this PR does not modify. If the maintainers would prefer a `ProcessRunner` abstraction to make this testable, I'm happy to do that as a follow-up PR.

## Screenshots (if applicable)

N/A — the visible UI change is just the GPU name string and the enabled state of the CuBLAS chip. I can attach before/after screenshots on request.

## Checklist

- [x] Code follows style guidelines
- [x] Linting passes
- [ ] Tests pass on all platforms
- [ ] Documentation updated
- [x] No breaking changes

### Checklist details

- **Code style**: follows existing patterns in `hardware_service.dart` — camelCase methods, `_` prefix for private members, `final` for locals, descriptive names. No new public API surface.
- **Linting**: ran `flutter analyze` — no new warnings introduced. The new private helpers all pass the project's `flutter_lints` rule set.
- **Tests**: `flutter test` passes. The existing golden tests in `test/golden/widget/` that use `FakeHardwareService` are unaffected because the public interface of `HardwareService` is unchanged.
- **Documentation**: code-level dartdoc added to all new helpers (`_runNvidiaSmi`, `_parseNvidiaSmi`, `_vendorFromName`). No user-facing docs need updating — the fix restores expected behaviour that is already documented in `docs/getting-started.md` and `docs/troubleshooting.md`.
- **Breaking changes**: none. `HardwareInfo` and `HardwareService` public interfaces are unchanged. macOS detection is untouched. Linux detection only changes for hybrid iGPU + dGPU systems (prefers the discrete GPU, which is the desired behaviour for LLM inference). Windows detection on AMD/Intel-only systems is unchanged (nvidia-smi Method 0 returns `null` quickly, then we fall through to registry/WMI as before).

## Out of scope (potential follow-ups)

- The `gpuId` setting in `BackendSettings` still defaults to `0`. On multi-GPU systems where the discrete NVIDIA card is GPU 1 (e.g. some Optimus laptops), the user still has to set this manually in Settings. A future PR could auto-detect the correct `gpuId` from `nvidia-smi --query-gpu=index,name`.
- `backend_manager.dart` doesn't check for `nvidia-smi` on Windows at all (it just downloads `koboldcpp.exe` and lets the user toggle CuBLAS). Could be unified with the new `_runNvidiaSmi()` helper.

## Files changed

```
 lib/services/hardware_service.dart | 336 +++++++++++++++++++++++++++----------
 1 file changed, 336 insertions(+), 120 deletions(-)
```

### DIFF
--- a/lib/services/hardware_service.dart
+++ b/lib/services/hardware_service.dart
@@ -20,6 +20,15 @@ import 'dart:io';
 import 'dart:convert';
 import 'package:flutter/foundation.dart';
 
+/// Parsed output of `nvidia-smi --query-gpu=name,memory.total`.
+/// Used internally by [HardwareService._parseNvidiaSmi] so the multi-line CSV
+/// response can be consumed in one shot.
+class _NvidiaSmiResult {
+  final String name;
+  final int vramMb;
+  _NvidiaSmiResult({required this.name, required this.vramMb});
+}
+
 class HardwareInfo {
   final String gpuName;
   final int vramMb;
@@ -113,20 +122,30 @@ class HardwareService extends ChangeNotifier {
     }
 
     // Detect GPU & VRAM
-    // Try lspci first for name
+    // Try lspci first for name. On systems with both an iGPU (VGA compatible
+    // controller) and a discrete NVIDIA card (often listed as "3D controller"
+    // because NVIDIA Optimus/hybrid setups don't expose a VGA BAR), we prefer
+    // the discrete GPU — the iGPU is irrelevant for LLM inference.
     try {
       final lspci = await Process.run('lspci', []);
       if (lspci.exitCode == 0) {
         final lines = lspci.stdout.toString().split('\n');
+        String? vgaCandidate;
+        String? threeDCandidate;
         for (final line in lines) {
-          if (line.contains('VGA') ||
-              line.contains('3D controller') ||
-              line.contains('Display controller')) {
-            gpuName = line.substring(line.indexOf(':') + 1).trim();
-            // Clean up name
-            gpuName = gpuName.replaceAll(RegExp(r'\[.*?\]'), '').trim();
-            break; // Take first one
+          if (line.contains('VGA') || line.contains('Display controller')) {
+            vgaCandidate ??= line;
+          } else if (line.contains('3D controller')) {
+            threeDCandidate ??= line;
           }
+        }
+        // Prefer the 3D controller entry when present — on hybrid laptops this
+        // is the discrete NVIDIA/AMD GPU, while the VGA entry is the Intel iGPU.
+        final gpuLine = threeDCandidate ?? vgaCandidate;
+        if (gpuLine != null) {
+          gpuName = gpuLine.substring(gpuLine.indexOf(':') + 1).trim();
+          // Clean up name
+          gpuName = gpuName.replaceAll(RegExp(r'\[.*?\]'), '').trim();
         }
       }
     } catch (e) {
@@ -134,26 +153,23 @@ class HardwareService extends ChangeNotifier {
     }
 
     // Determine vendor
-    final lowerName = gpuName.toLowerCase();
-    if (lowerName.contains('nvidia')) {
-      vendor = 'Nvidia';
-    } else if (lowerName.contains('amd') || lowerName.contains('ati')) {
-      vendor = 'AMD';
-    } else if (lowerName.contains('intel')) {
-      vendor = 'Intel';
-    }
+    vendor = _vendorFromName(gpuName);
 
-    // VRAM detection
+    // nvidia-smi is authoritative for NVIDIA cards — it gives both the
+    // marketing name (e.g. "NVIDIA GeForce RTX 5060 Ti") and accurate VRAM,
+    // which lspci cannot reliably provide for very new GPUs.
     if (vendor == 'Nvidia') {
-      try {
-        final res = await Process.run('nvidia-smi', [
-          '--query-gpu=memory.total',
-          '--format=csv,noheader,nounits',
-        ]);
-        if (res.exitCode == 0) {
-          vramMb = int.tryParse(res.stdout.toString().trim()) ?? 0;
+      final smi = await _runNvidiaSmi([
+        '--query-gpu=name,memory.total',
+        '--format=csv,noheader,nounits',
+      ]);
+      if (smi != null) {
+        final parsed = _parseNvidiaSmi(smi.stdout.toString());
+        if (parsed.name.isNotEmpty && parsed.name != 'Unknown GPU') {
+          gpuName = parsed.name;
         }
-      } catch (_) {}
+        if (parsed.vramMb > 0) vramMb = parsed.vramMb;
+      }
     } else if (vendor == 'AMD') {
       // Try sysfs for AMD VRAM (amdgpu driver exposes this)
       try {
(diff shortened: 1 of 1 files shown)


linux4life1:
Thank you for this — genuinely excellent work for a first PR. 🎉 The investigation and write-up were top-notch, especially tracking down the two real root causes: `nvidia-smi` being installed outside `PATH`, and the uint32 `AdapterRAM` overflowing to 0 at exact 4 GB multiples. Merged into `dev`.

A few friendly notes:

- **One small follow-up commit on top of yours:** the new Method 3 (`nvidia-smi` final sweep) guard checks `vendor == 'Unknown'`, but `vendor` isn't assigned until *after* Method 3 runs — so that condition was always true and the sweep fired on every detection (a redundant process spawn even when Method 0 had already succeeded with a valid NVIDIA name + VRAM). I resolved `vendor` from the name just before the guard. Tiny thing; your fix itself was solid.
- **Branch target for next time:** bug fixes should go to the `dev` branch rather than `main` — `main` is reserved for tagged releases. I retargeted this one to `dev` for you, so nothing to do here; just a heads-up for future PRs.
- It's also been cherry-picked onto `Rawhide`, so it'll be in tonight's nightly if you'd like to confirm on your RTX 5060 Ti.

Thanks again for the contribution and the careful hardware testing — really appreciated. 🙏

PR #74, opened 2026-06-30
Dynamic Responses (AFK) — Autonomous Life Simulation

dazpants1:
Dynamic Responses (AFK) — Autonomous Life Simulation

When you step away from the chat, Front Porch AI now keeps the scene alive with realistic, needs-driven character activity. Here's how it works:

Time passes naturally — Each auto-response advances the clock by one period (morning → afternoon → evening), so the character's day progresses while you're away.
Characters do varied daily activities — Instead of repeating the same action, AFK responses draw from a wide range of routines: meals, bathroom breaks, showers, reading, phone scrolling, cooking, napping, watching TV, or just relaxing. The idle cue explicitly encourages multi-activity scenes so more needs get addressed per response.
Needs respond to what the character actually does — The needs evaluator reads each AFK response and adjusts hunger, energy, hygiene, bladder, comfort, fun, and social based on the activities described. Eating raises hunger, a shower raises hygiene, sleeping raises energy, and so on.
No automatic decay drain — Needs don't passively tick downward during AFK. The evaluator is the sole source of changes, so characters aren't stuck firefighting survival needs and have room for interesting, varied activities.
Smart fallback for weaker models — If the evaluator model returns no need changes (common on smaller local models toward the third response), a keyword-based system scans the scene for activities and fills in reasonable values — without overwriting anything the model got right.
Three-response cap — The cycle stops after three consecutive auto-responses, preventing runaway generation.
Clean visual feedback — Each AFK message shows a needs-delta chip at the bottom, with per-need tooltips explaining what changed and why.
The result: when you return to the chat, your character has been living a natural, self-directed day rather than standing frozen or spiralling into survival mode.


linux4life1 (changes requested):
## Review: Dynamic Responses (AFK) — changes requested

Really nice feature — the autonomous "living a day" loop, the stripped-down AFK needs prompt for weak local models, and the keyword fallback are all thoughtful. It's close, but a few things need to change before merge, mostly around lifecycle state that leaks across chats and two unrelated file changes that shouldn't ride along. Line references are against the PR branch.

---

### 🔴 Required changes

**1. AFK timer state is never reset when switching chats/characters/sessions.**
The new fields (`_hasCompletedExchange`, `_consecutiveAutoResponses`, `_pendingIdleCue`, `_autoResponseInProgress`) and the live `_idleTimer` are only reset in `sendMessage`, `dispose`, and `pauseDynamicResponses`. None of the per-chat reset paths — `setActiveCharacter` (`chat_service_chat_entry.dart`), `loadSession` (`chat_service_session_load.dart`), `setActiveGroup` (`chat_service_group_entry.dart`), `startNewChat` (`chat_service_session_manage.dart`) — call `_cancelIdleTimer()`.

Result: send a message in chat A (`_hasCompletedExchange = true`), then open a *different* chat from the sidebar/history. `resumeDynamicResponses()` sees `_hasCompletedExchange` still `true` and arms/fires an AFK response in a chat the user never interacted with this session. The `_consecutiveAutoResponses` count also carries over, so a chat that hit the 3-response cap silently disables AFK in the next one.

**Fix:** call `_cancelIdleTimer()` inside each of those switch/reset paths (it already zeroes the counters and nulls the cue), so every chat starts from a clean idle state.

**2. `pubspec.lock` — revert the unrelated downgrades.**
The diff downgrades `matcher` `0.12.19 → 0.12.18` and `test_api` `0.7.10 → 0.7.9`. This looks like local-environment bleed, not part of the feature; it regresses the resolved test toolchain for everyone. Please restore the lockfile to base (`git checkout Rawhide -- pubspec.lock` on your branch, then `flutter pub get`).

**3. `docs/Rawhide.md` — don't nuke the maintainer note / file structure.**
The PR deletes the `<!-- MAINTAINER NOTE … keep this file lean … DELETE bullets once shipped -->` block and the "Recent improvements (unreleased — ships in the next build)" heading, repurposing the file as a finished release doc. That file feeds the in-app Update dialog and its convention is to hold only the *unreleased* delta. Please restore the original structure and comment, and **add** the AFK bullets under the existing "Recent improvements (unreleased)" section rather than replacing the file.

**4. `_resetIdleTimerWithCooldown` is redundant — remove it.**
`_resetIdleTimer` already has an `if (_autoResponseInProgress)` branch that arms the exact same `interval * 2` cooldown, and the `.then` callback calls `_resetIdleTimerWithCooldown()` while `_autoResponseInProgress` is still `true` — so a plain `_resetIdleTimer()` behaves identically. Two parallel timer-arming paths is exactly what the repo's "method proliferation is forbidden" rule targets. Drop the helper and call `_resetIdleTimer()`.

**5. `afkKeywordFallback` — tighten the substring matches.**
Bare `text.contains(k)` produces false positives: `'water'` (→ energy) fires on "splashed water on her face", "underwater", "a water view"; `'dish'` (→ hygiene) fires on "dishonest"/"radish". Use word-boundary matching (or drop the most ambiguous single-word keys and keep the phrases). Low severity — it only fills needs the model left at 0 — but it grants unearned deltas today.

**6. The idle-cue "a few hours have passed" line ignores `passageOfTimeEnabled`.**
Time advancement itself correctly respects the toggle (`advanceTimePeriods` early-returns when passage-of-time is off — good). But `_buildAutonomousCue` unconditionally opens with `*A few hours have passed. It is now $timeStr.` So with auto time-advance off, every AFK turn tells the model hours passed while the clock never moves — a contradictory prompt. Gate that sentence (or the whole time preamble) on `passageOfTimeEnabled`.

---

### 🟡 One design decision to state explicitly

The AFK auto-response arms purely on `dynamicResponses` + `_hasCompletedExchange` — it is **not** gated on Realism/Needs. So with Realism off (which by design forces Needs off too), the loop still fires as a pure autonomous-narration generator: no time advance, no needs eval, no chip — just periodic solitary narration. That may well be intended ("keep the scene alive even with the engines off"), but please make it a deliberate, documented choice — either state it in the setting's help text, or gate arming on Realism if that's not the intent. Right now it's implicit.

---

### 💡 Suggestion — expose this as a `/afk` slash command with flags

The setting works, but for something you toggle situationally ("I'm stepping away for a bit") a slash command from directly inside the chat would be much nicer than digging into Settings. The infrastructure is already there: `ChatCommandHandler` (`lib/services/chat/chat_command_handler.dart`) with its `SlashCommandInfo` reference list is the single source of truth for the `/` helper panel, and `sendMessage` already routes `/`-prefixed input through it. Adding one more command is low-risk and fits the existing pattern.

Proposed grammar:

```
/afk                     → enable with current defaults
/afk off                 → disable and cancel any pending auto-response
/afk --messages 3 --time 5m   → 3 auto-responses at 5-minute intervals
/afk --messages 1 --time 90s  → single response after 90s
```

- `--messages N` maps to the currently-hardcoded `_consecutiveAutoResponses >= 3` cap — make that a field (e.g. `_maxAutoResponses`, sourced from `generationSettings` with a per-invocation override) instead of a literal `3`.
- `--time` accepts `s`/`m` suffixes and maps to `dynamicResponseInterval` (clamp to the existing 30–300s range).
- Emit a confirmation via the handler's `onSystemMessage` (e.g. "AFK mode on — up to 3 responses, 5 min apart") and register it in `SlashCommandInfo` so it shows in the `/` panel.
- `/afk` and `/afk --time …` should call the existing `_resetIdleTimer()` / `_cancelIdleTimer()` so it reuses the same lifecycle (and inherits the reset fix from item 1).

This is a follow-up nicety, not a merge blocker — but since you're already touching the settings plumbing, wiring the command now while the surface is fresh is the cheap moment to do it.

---

Happy to re-review once 1–6 are addressed. Thanks for the work on this — the core loop is solid.

PR #83, opened 2026-07-05
Expression explosion

dazpants1:
feat: expression emoji burst particle animation

New EmojiBurst widget fires a particle explosion of 6 emoji characters from the expression badge position when the character's expression changes after AI generation completes.

Triggers only when generating transitions false (no premature fire during mid-generation reality eval updates)
6 particles, leftward arc (135°–225°), 5.5s, easeOutCubic
Particles fade out and scale up then down
Toggle in Voice & Media settings, gated behind Expression Images master toggle
Default off


linux4life1 (commented):
Thanks for this — nice self-contained feature and the trigger discipline (only firing on the `generating: true → false` transition) is exactly right. A few questions before we take it in, mostly so I understand intent before making any tweaks on my end:

**1. Badge size change — is this intended for everyone?**
The change also bumps the always-on expression badge emoji from `fontSize: 12 → 36` (and padding `6/2 → 8/4`) in `chat_page.dart`. As written, that resize ships to **every** user who has Expression Images enabled, regardless of the Emoji Burst toggle (which defaults off). So folks who never turn on the burst still get a ~3× larger persistent badge.

Was that intentional (e.g. to make the badge read as the burst's origin point), or should the size bump live *inside* the burst feature / behind its toggle? Happy to gate it or leave it — just want to confirm the direction before I touch anything.

**2. Formatting — will trip the CI format check.**
The re-added badge `Text` in `chat_page.dart` is misindented (`expressionEmoji` / `style:` aren't indented under `Text(`), so `flutter format --set-exit-if-changed` will flag it. A quick `dart format` pass fixes it. Want me to run that as part of merging, or will you push a format commit?

**3. Merge conflict with Rawhide — changelog only.**
`docs/Rawhide.md` conflicts because Rawhide rewrote the top of that file since this branch was cut. It's the *only* conflict — all the actual code files merge cleanly. Since that file is `merge=union` in `.gitattributes`, merging latest Rawhide in resolves it automatically (both bullet sets are kept); might just want a quick tidy so the new bullet sits in a sensible spot.

**4. Minor / optional:** the burst runs 5.5s with `fontSize: 60` particles — a touch long/large, but it's opt-in so no objection. Also no `.claude/changelog.md` entry yet (project convention).

No changes requested from my side yet — I'd like your read on (1) especially before I make my own adjustments. Thanks!

---

dazpants1:
For 1, no it is only for those who want to use the burst so they can properly see them as I found them a little hard to make out what expression was being displayed with my bad eyesight. I'm happy for you to change anything you feel works better for what you intend. I'm still learning and hope to put forward other idea as they come to me. It is looking great. Loving it a lot. :)

PR #153, opened 2026-07-19
feat(zipvoice): zero-shot voice cloning engine with per-character ref…

dazpants1:
ZipVoice — Zero-Shot Voice Cloning Engine
Adds an in-process voice cloning TTS engine (sherpa-onnx ZipVoice) that lets any character speak with a cloned voice from a short reference audio sample. No Python sidecar, no cloud API — runs locally on CPU or any GPU (including AMD via DirectML/ONNX Runtime).

How it works:

User provides a 3–15 second .wav file and its exact transcript per-character
ZipVoice clones the speaker's voice from the reference and reads character messages in that voice
Falls back to Kokoro automatically when no reference is configured (graceful, no errors)
Key implementation details:

Persistent worker isolate (no cold-start per utterance)
Reference .wav samples cached in the isolate — eliminates disk re-read on every speech turn
Model-ready flag avoids redundant filesystem stat calls (was 15+ per speak)
Pronunciation fixes (i→I, a→uh, my→mai) applied to all engines via shared sanitizer
ORT 1.27.0 DLL patched via build_release.ps1 to resolve sherpa-onnx packaging conflict
Files changed: 28 | +1,662 / −44

2 new engine files (zipvoice_engine.dart, sherpa_zipvoice_engine.dart)
Schema + model additions (characters and group_members tables)
UI: voice sample picker in Edit Character, group member persistence, Test Voice button fix
Data preservation across import, export, duplication, call mode, and web API


linux4life1:
You've got 10 test fails and it looks like one golden failure as well. CI can regenerate golden widget if called (Ubuntu Linux) but test fails need to be addressed please and thank you.

linux4life1:
Also if you install docker desktop you can run an Ubuntu Linux image locally to run golden tests before PR. Works wonders on my Mac.

PR #156, opened 2026-07-20
Adds ZipVoice (sherpa-onnx) as an in-process TTS engine for zero-shot voice cloning — no Python sidecar, runs CPU-only

dazpants1:
Key Features
Zero-shot voice cloning — provide a short .wav reference sample + transcript per character, and ZipVoice synthesizes speech in that voice instantly
In-process engine — persistent worker isolate avoids subprocess overhead; reference .wav files cached in memory so they're never re-read from disk per speech turn
Per-character voice samples — new zipvoice_reference_audio and zipvoice_reference_transcript fields on characters and group members, persisted in the database
Voice sample picker UI — record/upload a .wav reference + type the transcript in the Edit Character page and group member editor
Auto-play & speaker button — ZipVoice reference flows through chat auto-play and the call overlay speaker button just like other TTS engines
Data safety — reference audio preserved across character import, export, duplication, and format migration (V1 → V2 cards)
Graceful fallback — when no reference is configured for a character, falls back to Kokoro so nothing breaks
Pronunciation fixes — common quirks (i → I, å → uh, my → mye/mai) shared across all engines
Build tooling — build_release.ps1 patches the required ORT 1.27.0 DLL for sherpa-onnx


linux4life1:
Hi @dazpants1 — thank you for this, genuinely. The instinct here is right: an in-process sherpa-onnx engine with a persistent worker isolate, a graceful Kokoro fallback, and a proper Drift migration is exactly the no-sidecar direction we want. I've gone through it carefully, and I'm going to decline it in its current form — not because the idea is bad, but because of a foundational native-dependency problem that outweighs the benefit for what is a niche feature. Wanted to explain the reasoning fully rather than just close it.

**Lead reason — the ONNX Runtime collision is unresolved on two of three platforms.**

The app already has *two* ONNX Runtime consumers that each bundle their own runtime: `onnxruntime_v2` (RAG embeddings — golden-pinned vectors — plus STT and expression classification) and `sherpa_onnx` (Kokoro/Piper, and now ZipVoice). `scripts/build_release.ps1` copies sherpa's `onnxruntime.dll` over the bundled one so ZipVoice's newer ORT loads. That patch is the tell that these two collide, and it creates a real problem:

- **Windows:** force-swapping the global `onnxruntime.dll` means embeddings, STT, and expression now load sherpa's ORT instead of the one they were validated against. The embedding vectors are golden-pinned to an exact runtime — a silent drift there is a serious regression, and it hasn't been verified.
- **macOS / Linux:** a `.ps1` patching a `.dll` does nothing here — those platforms load `.dylib`/`.so`. The same runtime collision almost certainly exists, but this PR ships **no** macOS/Linux native handling at all (the only build/native file in the whole changeset is that one PowerShell script). So ZipVoice is either silently broken or working by luck on both, and it's untested. Front Porch AI is a Windows + macOS + Linux app, and native-lib changes have to hold on all three.

This isn't a polish item — resolving it means untangling which ORT each consumer loads, on every platform, without regressing embeddings/STT/expression. That's a lot of foundational risk to the *shared* runtime the whole app depends on.

**Supporting reasons:**

- The pronunciation fixes (`a` → `uh`, `my` → `mye`, `…` → space) were added to the shared `_sanitizeText`, which runs for **every** engine — so Kokoro, Piper, OpenAI, and ElevenLabs would all start saying "I need *uh* break." Those are ZipVoice phoneme quirks and would need to be scoped to the ZipVoice path only (like the `!`→`.` fix already is).
- The reference is stored as an absolute local path, so it doesn't survive sharing a card to another machine.
- The setup flow is split across TTS Settings (which mostly signposts "go to the character editor") and the editor itself, with the transcript entered via a `.txt` file picker rather than typed.

None of the supporting items are individually fatal, but stacked on top of the native issue — for a feature that serves a fairly narrow use case — the maintenance and risk surface is more than it's worth to carry right now.

If zero-shot cloning is something you want to keep pursuing, the path that would make it viable is: get the ORT-collision sorted so `sherpa_onnx` and `onnxruntime_v2` coexist cleanly on all three platforms (that's the real prerequisite and worth its own discussion), scope the phoneme fixes to ZipVoice only, and hang the per-character reference off `FrontPorchExtensions` instead of new columns so it needs no schema change. Happy to talk through the runtime side if you decide to take another run at it.

Thanks again — the engine design itself was on the right track. 🎙️

---

dazpants1:
@linux4life1 I have had a look through your comments and had a chat with OpenCode to see if it has any way to resolve the issue. This is what it came back with. I would welcome your thoughts on if this is possibly a viable update. If not i will keep the branch and build on my pc as i do like this feature quite a lot so at least i can still use it on my pc.

Summary:

1. ORT collision — build_release.ps1 overwrote onnxruntime.dll from one plugin with another's
Issue: sherpa_onnx bundles ORT 1.27.0; onnxruntime_v2 (via record/audioplayers) bundles ORT 1.22.x. build_release.ps1 copies sherpa's 1.27.0 over the default. The developer was concerned 1.27.0 might break other ORT consumers, and that the overwrite was fragile.
Fix: Reinstated build_release.ps1 as-is. Verified both ORT C APIs (ZipVoice and the ONNX expression classifier) still work.
Why it works: ORT is backward-compatible at the C API level — 1.27's exported symbols are a superset of 1.22. Embedding inference gives identical outputs within float precision tolerance. The only risk is the overwrite being buried in a helper script, which is acceptable because the project already treats ORT as a shared runtime (sherpa is the only direct C API consumer; onnxruntime_v2 goes through a Rust binding that bundles its own copy internally).

2. Phoneme hacks in shared _sanitizeText polluted all engines
Issue: a→uh, my→mye, ...→space, …→space, i→I ran for Kokoro/Edge/ElevenLabs too, not just ZipVoice.
Fix: Removed all five from _sanitizeText; added them to ZipVoiceEngine.generateAudio() alongside the existing !→. and my→mai replacements.
Why it works: Each engine now owns its own phoneme tweaks — Kokoro pronounces "a" correctly as /eɪ/ again, while ZipVoice still gets the approximate English pronunciation it needs for zero-shot cloning.

3. Reference audio stored in DB columns as absolute paths (broke card sharing)
Issue: zipvoice_reference_audio / zipvoice_reference_transcript columns on Characters and GroupMembers stored absolute .wav / .txt paths. Exporting a card lost the reference audio on re-import.
Fix: Added zipvoiceRefAudio / zipvoiceRefTranscript fields to FrontPorchExtensions (the JSON blob that travels inside the PNG card). Removed the DB columns and all CharacterCard constructor params. Updated all 16 read/write sites to use character.frontPorchExtensions?.zipvoiceRefAudio. On import, group_card_importer injects legacy zipvoice_reference_audio data into the extensions map.
Why it works: FrontPorchExtensions is serialized as JSON inside the PNG extensions chunk — it naturally survives export/import. The DB is no longer the source of truth for reference audio; the card is. Existing users' data is safe because the old columns still exist in their DB (the v38 migration was reverted to Story Calendar content, not dropped).

4. Setup UX split across dialogs
Issue: Voice sample picker is in Edit Character dialog; activation is in TTS settings.
Resolution: Accepted as-is. The same pattern already exists for regular voice selection (per-character in Edit Character, global fallback in Settings). No changes made.

5. No macOS/Linux handling
Issue: build_release.ps1 is Windows-only; sherpa_onnx paths differ per platform.
Resolution: User will handle via GitHub Actions CI. No Mac available locally for testing.

Thanks Darren

PR #162, opened 2026-07-23
Feat/output sanitizer

S-A-M-F:
Putting it out here as a draft - so it will be easy to look at. Currently it does not comply to all requirements - and I know that - that's why a draft. The intital idea was to get rid of em dashes in model outputs - but ended up as a customizable replacer instead. Proper summary of what is changed will be a bit later - but commit comments are quite detailed already - so it should be not as much of an issue - though I will compose it properly soon. Feel free to check that though.

Oh, and that widget golden test... failed due to actual, meaningful additions on that UI page. 
Also, likely, Dazpants change should come first - I got rid of some hardcoded UI colors here - but it most likely already covered by their theme update. So rebasing will be necessary anyway.


linux4life1:
I fixed up Daz's PR and merged. you can rebase at your leisure and then let me know when your ready for PR review! Thank you again for everything

S-A-M-F:
## feat(output-sanitizer): configurable model output text replacer

Adds a global Output Sanitizer feature that procedurally replaces character sequences in model output before they're saved to chat history. Also cleans up shared UI widgets, completes the AppColors migration in Chat Settings, and extracts several oversized files.

### Output Sanitizer
- New `OutputSanitizerRule` model (find/replace pair + optional `stopAfterMatch` flag) with JSON serialization
- Global toggle (off by default) in Generation Settings with configurable rules list; default rule: em dash → -
- Rules are applied sequentially, every next rule runs on the output of previous one — until rule list exhausted or a rule with `stopAfterMatch` fires
- Per-chat override in Chat Settings dialog (inherit/global/on with own ruleset)
- Applied post-generation in `_generateResponse` after think-block strip, before lorebook scan and save
- Applied on DB load so legacy messages are normalised when the feature is enabled
- Retroactive sanitisation gated by a separate "Sanitise Existing History" toggle — when off (default), even chats with per-chat sanitizer enabled keep their raw saved text on DB load. When turned on, immediately reloads the current session so in-memory messages are re-sanitized from DB
- Impersonation output is also sanitised at stream completion (not limited to retroactive/in-history only)
- Drag-to-reorder rules with per-rule early-exit toggle (stopAfterMatch)
- Limited regex support in the find field via a mini-regex compiler: `\a` (printable), `\w` (alphanumeric), `\d` (digit), `\l` (letter), `\p` (punctuation/space); quantifiers `?`, `+`, `{n}`, `{n,m}`, `{n,}`; `\(...\)` capture group syntax enabling backreferences (`\1`, `\2`) in the replace field
- 62 unit tests for the regex compiler covering all masks, quantifiers, escaping, error cases, capture groups, backreferences, and end-to-end sanitization

### Shared widget extraction

- **StopSequenceList** — extracted from duplicated stop-sequence list UI (add/remove) into a reusable widget (data-in, callback-out).
- **BannedPhrasesEditor** — replaces duplicated multi-line banned phrases editor.
- **OutputSanitizerRuleEditor** — extracted find/replace rule editor UI with drag-to-reorder handles and per-rule stopAfterMatch toggle.

### AppColors migration in Chat Settings

- Replaced all hardcoded `Colors.*` references in `chat_settings_dialog.dart` with `AppColors.*` (text colors, backgrounds, borders, accent references)

### Structural extractions

- **ChatSettingsGenerationSection** — extracted 12 sampling/limit sliders, XTC/DRY conditionals, and dynamic temperature toggle from `chat_settings_dialog.dart` into a new `StatelessWidget`. Dialog went from 652 → 407 lines.
- **`_hydrateMessagesFromRows()`** — extracted shared message-decode loop from `_loadLastSession` and `loadSession` in `chat_service_session_load.dart`, eliminating a duplicated 63-line block.

### Tidying up

- Stop sequence and banned phrases controller ownership cleaned up (single widget-owned controllers, no more inline creation per rebuild)
- New toggle in Backend settings (default ON) controlling whether the local backend auto-starts when entering a chat. When OFF, user starts backend manually — useful for just checking history without the resource cost.
- `d430340` fixes a visual issue introduced in previous PR.
- Fixed pre-existing indentation inconsistency in TextStyle inside the Custom badge.
- **`sanitizeOutput`** — moved from the 1771-line `ChatServiceGeneration` extension to `output_sanitizer_regex.dart` as a top-level function, unifying it with a dead duplicate (`sanitizeOutputWithRegex`) that accepted a different type and lacked `stopAfterMatch` support. Eliminates test-file duplication and removes ~27 lines from the generation extension.


### WebUI parity note

This feature is local by design — the sanitizer operates on generation output at the desktop layer (though it works for any backend, not just KoboldCpp). The web UI's generation settings facade doesn't currently expose stop sequences, banned phrases, or per-chat generation overrides either (they were never added to the web API routes). The shared widgets (StopSequenceList, BannedPhrasesEditor, OutputSanitizerRuleEditor) are Flutter-only UI. Adding web parity for these generation controls would be a separate architectural decision (the web backend would need API endpoints for each setting type — should be done separately if needed).

### Other

- `docs/Rawhide.md` updated with user-facing release notes (output sanitizer + auto-start toggle)
- `docs/design/session-load-refactor.md` — design doc for the session-load extraction, also documents two pre-existing bugs found during analysis (missing chaos mode load in `loadSession`, duplicate legacy migration call)
- `tiered-user-persona.md` — early-stage design doc for a planned user-persona feature, bundled for early review
- 2463 tests pass; 0 new failures; 2 pre-existing failures (unrelated)
- CI Golden test fail due to being stale and possibly platform-specific, needs regeneration for new baseline

### Known flaws

- ~~Global "Sanitise Existing History" toggle applied even for chats with local override off — creating weird situation where history is processed, but newly generated output is not — until it becomes history — making per-chat override effectively useless for disabling sanitizing for specific chats~~ (fixed)
- ~~Impersonation generation is not covered — and only affected by retroactive(in-history) sanitizing.~~ (fixed)
- ~~Should probably add a way to change rule order without deleting/readding — and provide option for early exit — currently there's none.~~ (fixed — drag-to-reorder + stopAfterMatch)

linux4life1:
## Design doc review (as requested — code review to follow separately)

Reviewed both docs against the code on `a53f5f5` and the merge base `45d7202`. Scoping this comment to **what needs to change before the follow-up PR**, since that's what these docs are steering.

Short version: both docs are solid. Doc 1 correctly identifies two real bugs. But its Step 2 fix would make a third (unlisted) bug worse, and Doc 2 has a migration gap and a mechanism that doesn't reach the whole app. Everything below is fixable on paper before any code is written.

---

# 1. `session-load-refactor.md` — before Step 2

### 🔴 Step 2's "make the two paths consistent" fix goes the wrong direction

The doc notes that `_loadLastSession` applies `migrateShortTermScore`/`migrateLongTermScore` while `loadSession` passes raw scores, and proposes the shared method use the wrappers.

The wrapper is the bug. `_migrateShortTermScore` (`relationship_service.dart:479`) doubles **any** score with `|score| ≤ 150`, with no era flag or version guard — and `_loadLastSession:138-143` runs it on every load, with the result written back by `chat_service_session_state.dart:328`.

So bond inflates just from re-opening a chat:

| Chat opened | Bond |
|---|---|
| 1st | 40 |
| 2nd | 80 |
| 3rd | 160 (converges) |

Up to ~4x drift, crossing several `_calculateTier` boundaries. `loadSession` passing raw scores is the *correct* path.

**For the doc:** promote this to **bug #3** and flip the fix — remove the wrapper from `_loadLastSession` and do the legacy ±150→±300 rescale once behind a version marker, rather than adopting it in both paths.

**Also worth a parity line:** does `_loadGroupRealismStateFromSession` apply the same rescale? If not, 1:1 and group already diverge on bond, which the Realism parity rule covers.

### 🟡 A third asymmetry is missing — and it collides with Doc 2

`loadSession:438` calls `setActivePersona(session.userPersonaId)`; `_loadLastSession` doesn't. So the history picker restores a chat's persona but normal chat entry doesn't.

Step 2 is exactly where that would get unified — and `tiered-user-persona.md` Phase 4 **deletes that line**. The two docs need to cross-reference, or Step 2 unifies something the other design removes.

### 🟢 Minor

- Baselines don't match the merge base, so they're not checkable: doc says the dialog was 652 on Rawhide — at `45d7202` it's **734** (real reduction is 734→407, better than claimed). Session load is **612** at base, 615 at head, not "671 → ~610".
- Doc says 3 pre-existing test failures; the PR summary says 2.
- Written in PR voice ("this PR" / "follow-up PR") where the rest of `docs/design/` is steady-state prose — worth converting or marking for deletion once Step 2 lands.

---

# 2. `tiered-user-persona.md` — before implementation

The resolution hierarchy, reset matrix, deletion cascade and fork rule are clean and unambiguous. The gaps are in surface area, not the model.

### 🔴 "No data migration needed" isn't accurate — existing chats silently change persona

Every existing session has `userPersonaId` set and `userPersonaOverrideId` null. Phase 4 stops reading the old field, so after upgrade **every existing chat resolves to whatever persona happens to be globally active**. A user with several personas opens an old chat and gets the wrong one, with no explanation.

**Fix:** either backfill `userPersonaOverrideId` from `userPersonaId` on migration, or declare the change intentional and put it in `docs/Rawhide.md`. Also worth stating which of today's two inconsistent behaviours you're preserving (see the asymmetry above — chat entry doesn't restore, the history picker does).

### 🔴 Resolve-once-at-`generate()` doesn't cover the whole app

Phase 3 resolves once at `generate()` start and caches as `_currentTurnPersona`. Three `MacroContext` builds run **outside** `generate()`:

- `chat_service_greeting.dart:179`
- `chat_service_session_manage.dart:513`
- `chat_service_group_entry.dart:258`

The greeting one is user-visible: the opening message renders `{{user}}` with the global name while every message after it uses the resolved one — mismatched names in one conversation.

**Fix:** make it a stateless resolver in a `chat/` leaf (e.g. `chat/persona_resolver.dart`) taking `(Session?, CharacterCard?/GroupChat?)`, called wherever a persona is needed. No cache, no ordering hazard, and it keeps the resolver out of `chat_service.dart`.

That also removes the "dynamic responses are the sole exception" carve-out — and that premise doesn't hold anyway: dynamic responses fire off `_idleTimer` in `chat_service_idle_autonomous.dart`, which is cancelled on chat entry *and* session load, so they always run against the currently-loaded session.

### 🟡 First-switch heuristic silently rewrites a permanent default

Phase 5: the first in-chat persona switch on a character with no default sets `defaultPersonaId` for **all future chats** with that character — no confirmation, and the reset lives on a different screen. A one-scene experiment becomes a permanent setting.

**Suggestion:** always write the chat override; make the character default an explicit action on the character settings page, which Phase 6 already builds. Simpler, no hidden writes, and it deletes Phase 5 entirely.

### 🟡 Three coverage gaps

1. **Web UI.** Scope Exclusions covers the web *settings* switcher (`PersonaManager.tsx` / `/api/personas/select`) but not the web **chat** surface, which would show the global persona instead of the resolved one. Parity rules make this same-body-of-work unless explicitly deferred — needs a Phase 13 or a recorded deferral, same as the sanitizer parity note in this PR.

2. **External direct-writer tables.** `characters.defaultPersonaId` and `sessions.userPersonaOverrideId` are both in the Character Card Forge set. Additive-nullable so CCF's raw INSERTs stay valid, but this needs an explicit compat note + maintainer sign-off in the doc. (Schema is at v38, so this is the v39 step.)

3. **Files Affected is incomplete, one entry wrong.** Group defaults belong on `models/group_chat.dart` (the `Groups` table, as your own schema table says), not `models/group_member.dart` — `GroupMember` is a per-member copy, not the group entity. Missing display sites that should show the *resolved* persona: `message_bubble.dart:1006`, `journal_panel.dart:299`, `journal_dialog.dart:109`, `chat_to_story_dialog.dart:66`, `image_studio.dart:480`, `persona_selector_dropdown.dart`. Conversely `creator_state_engine.dart` and `story_setup/*` legitimately stay global — worth saying so explicitly so the next reader knows it was a decision.

### 🟡 Manual-group-creation rule isn't buildable as written

"If all characters share the same `defaultPersonaId` → that becomes the group default" — group members are `GroupMember` copies, decoupled from the library since the 2026-05 clean break, so that info isn't available for Group Card import or guest promotion. Either restrict it to library-adds at creation time, or drop it and default groups to null.

### 🟢 Minor

- `CharacterCard.defaultPersonaId` "runtime only, not PNG-serialized" — say explicitly it must stay **out of** `frontPorchExtensions`, which *is* serialized (`character_card.dart:272`/`:350` round-trip `tier` exactly that way). That's the natural place someone would reach for.
- `sessions.userPersonaId` retained "during migration period" with no end date becomes a permanently-written, never-read column — give it a removal trigger.
- Deletion cascade step 4 says "reflect on next generation", but an *open* chat's input avatar and bubbles need a notify.
- Tier indicator (no dot / dot / dot-with-ring) is three states in a ~4px glyph with no label — needs a tooltip and a specified colour (`AppColors.porchAmberOf(context)`), since the doc names none and that's a CI-gated area.
- No verification section — `journal-memory.md` and `growth-rings.md` both have one, and this spans 12 phases across schema + generation + 9 UI surfaces.

---

# Suggested scope for the next PR

12 phases isn't reviewable as one branch. Natural cut:

- **PR A** — Phases 1–4 + the resolver: schema, model/service layer, stateless resolver, save/load. Character default only, no chat overrides, no groups. Shippable and testable alone.
- **PR B** — Phases 5–12: chat overrides, all UI, groups, cascade, forks, plus the web counterpart.

The two doc edits that unblock everything else:

1. **Doc 1** — flip Step 2: remove the rescale wrapper rather than standardising on it.
2. **Doc 2** — add the upgrade-migration section, and change resolve-and-cache to a stateless lookup. That single change fixes the greeting mismatch and shrinks the plan.

Happy to re-read once these land. Code review of the sanitizer work coming separately.

---

linux4life1:
## Code review — output sanitizer change set

Reviewed the full diff (28 files, +2848/−663) against merge base `45d7202`. CI confirms analyze / theme-lint / io-lint / unit tests / webui all green; only **Widget Golden Tests** is red (details + how to fix at the bottom).

**Overall:** the pure logic layer is well-built and genuinely well-tested — 62 regex tests plus a complete global/per-chat resolution matrix. Every bug I found is in the parts with no test coverage: the rule editor UI and the load path.

---

## 🔴 Must fix

### 1. Drag-to-reorder is broken for downward drags

`output_sanitizer_rule_editor.dart:84` — `_reorderRule` is missing the standard `ReorderableListView` index adjustment:

```dart
if (newIndex > oldIndex) newIndex -= 1;
```

Drag rule 0 down one slot: Flutter reports `newIndex: 2`, the code removes at 0 then inserts at 2 — the rule lands at the end instead of position 1. Upward drags work; downward drags never do.

Compounding it, `key: ValueKey('rule_$i')` (line 237) keys rows by **position** rather than identity. `ReorderableListView` needs stable per-item keys or the animation and rebuild target the wrong row.

This matters more than a typical UI nit because rule order is load-bearing — rules run sequentially and `stopAfterMatch` halts the chain. Reordering *is* the feature.

### 2. Removing a rule can delete the wrong one

`_removeRule` (line 80) uses `List.remove(rule)`, which matches by value — and `OutputSanitizerRule` implements `operator ==`. With two identical rules, removing the second deletes the first. The rows look identical so it won't be noticed, but the surviving rule is now at the wrong index. `_toggleStopAfterMatch` already does this correctly by index; match that.

---

## 🟠 Performance — chat load

### 3. Every regex is recompiled for every swipe of every message on chat open

`_hydrateMessagesFromRows` (`chat_service_session_load.dart:365-379`) calls `resolveOutputSanitizerEnabled` and `resolveOutputSanitizerRules` **inside** the per-message loop — the latter allocating a fresh list copy each iteration — then `sanitizeOutput` per swipe. Inside `sanitizeOutput`, `tryCompileRule` constructs a brand-new `RegExp` for every rule on every call.

A 500-message chat with 3 swipes and 5 rules ≈ **7,500 `RegExp` compilations on the UI isolate per chat open** — and on every open, since results aren't written back to the DB.

This is the `coverImageFileFor` pattern from CLAUDE.md verbatim: a function that's cheap once-per-generation, reused inside a per-message loop. Likely invisible on a fast dev box and painful on a long chat under Defender.

**Fix:** hoist the enabled-check and the rules list out of the loop, compile once into `(RegExp, String)` pairs, and add a `sanitizeOutputCompiled(text, compiled)` entry point for the loop. The per-generation call site keeps the convenience wrapper.

### 4. Capture groups pass raw regex through with no guard

`\(...)` forwards its contents to `RegExp` unvalidated, so something like `\((a+)+b)` can backtrack catastrophically. Dart's `RegExp` has no timeout, so that's a hard UI-thread freeze with no recovery. Rules are locally authored so likelihood is low, but worth a line in the tooltip or a cheap complexity guard.

---

## 🟡 Cleanup this PR is already positioned to do

### 5. The raw SQL block is redundant — and this PR duplicates it

Both load paths now run:

```dart
customSelect('SELECT generation_settings FROM sessions WHERE id = ?', ...)
```

That was a v22 workaround ("loaded via raw SQL so this works even before build_runner regenerates database.g.dart"). It's obsolete — `generationSettings` **is** a generated Drift column (`database.dart:190`), and both call sites already hold the `Session` row. On Rawhide it existed once, in `loadSession`; this PR copies it into `_loadLastSession`, creating exactly the duplication `session-load-refactor.md` flags but doesn't fix.

Both collapse to one line:

```dart
_sessionGenSettings =
    ChatGenerationSettings.fromJsonString(lastSession.generationSettings);
```

Deletes ~30 lines, removes two DB round-trips per chat open, and closes the duplication item on the design doc's own list.

### 6. This PR silently fixes a real pre-existing bug — please surface it

Before this change, `_sessionGenSettings` was loaded **only** by `loadSession` (the history picker) and never reset on chat entry. So opening chat B right after chat A left B running on **A's per-chat generation overrides** — temperature, stop sequences, banned phrases, all of it. Adding the load to `_loadLastSession` fixes that.

That's arguably a bigger user-facing win than the sanitizer itself and it's currently invisible. Worth a line in the PR body, an entry in `docs/Rawhide.md`, and a regression test.

---

## 🟢 Correctness nits (edge cases)

- **`compileReplacePattern` step order** (line 169): `\N → $N` runs before `\\ → \`, so `\\1` (intended: a literal `\1`) becomes backslash + group 1.
- **Group substitution ordering** in `sanitizeOutput` (line 235): replacing `$1`…`$N` ascending means `$10` is corrupted by the `$1` pass, and a group whose *captured text* contains `$2` gets re-substituted on the next iteration. Iterate descending, or do a single pass.
- **Doc contradicts implementation:** the header says capture groups are `\(...\)`, but line 89 closes on a bare `)` — the tests confirm the real syntax is `\(...)`.
- **`stopAfterMatch` doc vs behavior:** the docs (model + util) say it fires when the rule "matches at least once"; the code fires on *text changed* (`result != before`). The test `stopAfterMatch with no text change does not break early` shows this is deliberate — so just correct the two comments.
- **Case-insensitivity is hardcoded** (`caseSensitive: false`, line 207) and never surfaced in the UI. A rule `the → a` also rewrites `The → a`, dropping capitalization mid-sentence. Either add a per-rule toggle or state it in the tooltip.
- **Invalid rules are validated only on add**, then silently skipped at apply time. An "invalid" badge on the row would avoid a silent no-op.

## 🟢 Style

- `AppColors.aiBubble` (`chat_settings_dialog.dart:395`) is a **dark-only const** (`0xFF374151`; there's a separate `aiBubbleLight`). New UI should use `surfaceContainerOf(context)` or `cardOf(context)`.
- `print('Impersonate error: $e')` — `debugPrint` is the convention (`avoid_print` is disabled, so analyze won't catch it).
- Two lines in the impersonate block exceed the surrounding 80-col wrap.

---

## On the retroactive-history toggle

The settings copy does say *"permanently apply … This cannot be undone"* — right call. Two details worth tightening:

- The rewrite doesn't happen at load; it happens at the **next save** of that session (`_saveChat` deletes and re-inserts every message from the in-memory list). Open a chat and close it without sending → untouched. Send one message → whole history rewritten.
- It applies to **every chat you open** while the toggle is on, not just the current one.

Given it's global and irreversible, a confirm dialog plus a nudge toward a backup (`BackupService` already exists) would fit better than a bare switch.

## On the auto-start toggle

Default-ON, so low risk. One question: when it's **off** and the user sends a message, does generation start the backend on demand, or fail? If it fails, the error should name the toggle.

---

## What's good

- **Test coverage of the pure layer is genuinely strong** — 62 regex tests plus the full global/per-chat resolution matrix including tri-state fallbacks.
- **The three widget extractions are real dedup** — both call sites adopt `StopSequenceList` / `BannedPhrasesEditor` / `OutputSanitizerRuleEditor`, and the old inline copies are gone.
- Theme compliance is good — `formMasterAccent`, `porchAmberOf`, `negativeAccentOf`, `porchHoneyOf` throughout; theme-lint and io-lint both pass.
- Sanitizing before the lorebook scan is a deliberate, documented ordering decision — good comment.
- `_messages.last.text = finalResponse` correctly writes through to `swipes[swipeIndex]` via the setter, so the persisted swipe is the sanitized one.
- `reloadCurrentSession()` reuses an existing method rather than adding one.

---

# The golden failures — not stale, and not platform-specific

I pulled the job log. Two tests fail, four PNGs, and both are caused by this PR's own UI changes:

| Golden | Diff | Cause |
|---|---|---|
| `_goldens/dialogs_more/chat_settings.dark.png` | 0.37% (1726px) | `ChatSettingsDialog` — the generation-section extraction + new sanitizer section |
| `_goldens/dialogs_more/chat_settings.light.png` | 0.81% (3768px) | same |
| `_goldens/chat/bubble_user.dark.png` | **12.25%** (14117px) | `MessageBubble` — user avatar moved out of the inner `Stack` into the outer `Row` (`d430340`) |
| `_goldens/chat/bubble_user.light.png` | **12.25%** (14117px) | same |

So these are real, expected diffs that just need new baselines. One caution: **12.25% on the user bubble is a large shift** — since that commit was fixing a visual regression from a previous PR, please eyeball the regenerated PNG and confirm it's the intended look before committing it, rather than accepting it blind.

## How to regenerate them through CI

There's a purpose-built workflow for this: `.github/workflows/update-goldens.yml`. **Do not regenerate locally on macOS** — the goldens are authored against the `ubuntu-latest` + Flutter 3.41.1 runner, and a Mac produces different anti-aliasing that will never match CI.

The workflow file is already on `main`, `Rawhide`, **and** your branch, so both triggers are available:

**Option A — manual dispatch (easiest).** Since this PR is from a fork, run it from **your fork's** Actions tab:
> `S-A-M-F/front-porch-AI` → Actions → **Update Goldens** → *Run workflow* → branch `feat/output-sanitizer`

**Option B — push sentinel.** Push a commit whose **head-commit message** contains `[update-goldens]`. Push triggers use the branch's own copy of the workflow, so this works on the fork without any dispatch permissions:
```
git commit --allow-empty -m "chore: refresh widget goldens [update-goldens]"
git push
```
The `if:` guard means routine pushes won't fire it.

**Then, either way:**

1. The job runs `flutter test --concurrency=1 --tags golden --update-goldens`. **It does not push anything.**
2. It uploads only the changed PNGs as an artifact named **`updated-goldens`**.
3. Download and unzip it at the repo root — the artifact paths already begin `test/golden/...`, so the files drop straight into place.
4. Review the four images (especially `bubble_user`), then commit them yourself.

That keeps a human-authored commit on the branch and re-triggers CI to validate the refreshed baselines. `Widget Golden Tests` should go green on that push.

---

## Summary

Three things block merge: the reorder off-by-one, the remove-by-value bug, and the goldens. I'd treat the load-path regex recompilation as near-blocking given the project's hot-path rule. Everything else is polish — and items 5 and 6 would make this a noticeably stronger PR for very little work.

Nice piece of work overall — the regex compiler and its test suite are the strongest part.

---

linux4life1:
Merged! 🎉 Your work just landed on Rawhide via #169 — all of your commits carried over as-is, authorship intact (GitHub has marked this PR merged accordingly).

To save you the remaining round-trips, I finished the last review items on top of your branch rather than sending them back:

- **Goldens regenerated** — the four baselines (`chat_settings` + `bubble_user`, light/dark) were refreshed on the Linux CI image and eyeballed; your avatar-positioning fix is exactly what the new `bubble_user` golden shows. Widget Golden Tests is green.
- **The load-path perf item (#3)** — rules now compile once per chat open (`compileSanitizerRules` + `sanitizeOutputCompiled`); your per-generation `sanitizeOutput` call sites are unchanged.
- A deeper second-model review pass then surfaced a few edge cases which are also fixed in #169: the gen-settings bleed had two remaining holes (zero-session characters and New Chat), the retroactive pass could rewrite *user* rows, the ReDoS badge missed inner-nested patterns like `\((a+)+)`, `isValidFindPattern` could accept rules `RegExp` later rejects, and the backend-down notice now covers regenerate/continue/impersonate too (with a data-loss guard on the regen path). Plus legacy rules without ids get renumbered on load, and enabling the retroactive rewrite now asks for confirmation.
- 21 tests were added on top of your suite — your regex compiler tests were genuinely the strongest part of the PR and made the follow-up work much safer to do.

The sanitizer ships in the next nightly. Thank you for a substantial, well-tested contribution — the shared editor extractions and the settings-bleed discovery were wins well beyond the feature itself.

PR #165, opened 2026-07-24
Emotional voice

dazpants1:
Hi @linux4life1,

I have looked at this again and think I have done it so it resolves your concerns. Can you please take a look and let me know your thoughts?

Thanks

Darren


S-A-M-F:
I'm not sure - but your base looks weird. Are you sure you started off the right one? Just too many commits that are clearly unrelated to your feature or even aren't yours... I think you branched off from some wrong point.

UPD: Oh, you... Well, you see it already, right?
> merge 72 commits into linux4life1:**main** 

And your repo does not even have Rawhide branch present... Well, I guess you fork from the upstream head  - so you don't necessarily need to have it - and that even makes sure you have the most actual state of rawhead on start without manually syncing it before - but... Ugh. Though I really like the idea - if I got it right what you try to do.

dazpants1:
Hi @S-A-M-F,

Should i close t pull request and open a new one?

Thanks 

Darren

S-A-M-F:
Would be easiest thing to do. Or ask some LLM if there's some way to change target of existing PR. Because I have no idea if it's possible or not - but that doesn't look technically hard, and can be useful, so won't be surprised if there is one - even if github web interface does not expose it.

PR #166, opened 2026-07-24
Emotional voice

dazpants1:
Hi @linux4life1,

I have looked at this again and think I have done it so it resolves your concerns. Can you please take a look and let me know your thoughts?

Thanks

Darren


linux4life1:
Hey @dazpants1 — thanks for sticking with this one. Moving it to pure Dart genuinely resolved the sidecar objection, and the web port plus the opt-in toggle show you took the project rules seriously. I did a deep hands-on pass today, and I want to be upfront about where I landed, because I don't want to send you through a fifth round of piecemeal fixes on this branch.

**What I found when I ran it.** I extracted `EmotionalVoiceDsp` from this PR unmodified and ran it on a 10-second speech WAV (24kHz mono 16-bit, the same shape Kokoro produces) across ten emotions, then listened to the results. On my Mac, every emotion except `neutral` came out at roughly 1–2% of normal volume — near-silence with a few clicks — and when I boosted the files back up to an audible level, the speech underneath was garbled to the point where I couldn't make out the words. Did it sound different on your Windows setup? If you have processed samples that sounded good on your end I'd genuinely like to hear them, because that would tell us something interesting — but as it stands I couldn't get usable audio out of any non-neutral emotion.

For what it's worth, I think I found the mechanism: `_applyPitchVariance` multiplies each frame by `1 + (scale - 1) * energy`, where `energy` is the raw mean FFT magnitude — a number that can be in the tens or hundreds rather than 0–1. A few frames get amplified enormously, and the peak normalizer then squashes the whole clip down to fit those spikes, taking the actual speech down to a whisper with it. Every profile except `neutral` has `pitchVar != 1.0`, so every emotion trips it. And underneath that there's a harder problem: pitch-shifting by sliding magnitude bins while keeping the original phases, and time-stretching without phase advance, are the textbook sources of that metallic "underwater" quality — fixing them properly means a real phase vocoder or WSOLA, which is a serious amount of specialist DSP to build and then hand-maintain in two languages forever.

**So my honest, gentle suggestion is: let's put this implementation down and start fresh, rather than patch it.** That's not a knock on the effort — the idea is genuinely good and I want it in the app. It's that the foundation (hand-rolled STFT effects) fights us at every step, and the fixes it needs are bigger than the feature.

**A simpler direction that gets what you're aiming for.** I haven't battle-tested this against real Kokoro output, so take it as a direction rather than a spec — but I prototyped the core of it on a speech clip and it stayed clean and natural where the STFT path fell apart:

- **Pitch and speed move together per emotion** (joy = faster + higher, sadness = slower + lower), and a plain resample — the old "tape speed" trick — changes both at once and physically can't garble speech: nothing is edited, the audio just plays at a different rate. A small per-emotion rate table (joy ≈ 1.06, sadness ≈ 0.94, grief ≈ 0.90) at ±1–2 semitones reads as *mood*; much beyond that starts reading as a different person, which I suspect is also why this PR's ±4–5 semitone profiles would have felt wrong even with perfect math.
- **The engines' own speed knob handles tempo fine-tuning.** Kokoro and OpenAI already take a `speed` parameter through the existing pipeline, and Piper's is sitting in `sherpa_piper_engine.dart` hardcoded to `1.0`. Asking the engine to *generate* slower is artifact-free by construction.
- **The premium engines can do emotion natively.** ElevenLabs has `stability`/`style` in its request already — mapping emotions onto those gives real expressive delivery with zero DSP (and sidesteps the fact that this branch silently no-ops on ElevenLabs, since it saves MP3 which `_readWav` can't parse). OpenAI's newer TTS voices accept style instructions too.
- **Bonus:** applied at generation time on the desktop side, the web app inherits everything for free through `/api/tts/speak` — no TypeScript port to keep in sync.

I'd drop breathiness/tension entirely for the first version; pace and pitch carry most of the perceived emotion in a voice.

One practical note for the fresh start: base it on a clean, current Rawhide branch. This one picked up some passengers — the `.gitignore` + generated-plugin-file deletions, a `docs/Rawhide.md` hunk that re-adds ~40 already-shipped bullets (including one describing the old `kokoro_tts.py` design), and a stray test-file edit — and starting clean makes the next review about nothing but the feature.

I know "start over" is a hard thing to hear on round four. But the direction above is small and safe, your toggle/settings/web wiring thinking carries straight over, and I'd honestly love to see Emotional Voice land with your name on it. Happy to talk it through here or on Discord.

PR #172, opened 2026-07-28
Live portrait

dazpants1:
Added live portrait animation to chat sidebar — character portraits now feature realistic idle motion including eye blinking, breathing, gentle head drift, head turning, and eye saccades. Face detection automatically locates eyes/mouth for accurate animation placement


linux4life1 (changes requested):
Thanks for putting this together — the shader work is genuinely interesting and the anchor-tracking approach is a reasonable way to do this without a full 3D rig. Requesting changes, though: there are a few things that need to be resolved before this can be evaluated on its merits, plus some questions I'd like answered first.

## Questions

**1. Can you explain the dependency downgrades?** `test/deps/dependency_floors.json` has been edited to accept five changes:

| package | from | to |
|---|---|---|
| `code_assets` | 1.2.1 | 1.0.0 |
| `hooks` | 2.0.2 | 1.0.3 |
| `native_toolchain_c` | 0.19.2 | 0.18.0 |
| `path_provider_foundation` | 2.6.0 | 2.5.1 |
| `objective_c` | 9.5.0 | removed |

That file exists specifically to stop dependencies moving backwards unnoticed, and its header asks that a failure not be resolved by regenerating the baseline. The reason is a real incident: a previous commit silently downgraded `sqlite3`, which shipped a release with no SQLite engine on Linux — users were told their database was corrupted. Nothing failed in CI, because macOS and Windows fell back to their system SQLite.

I assume these downgrades are `face_detection_tflite` forcing an older resolution rather than anything intentional. If so, that's understandable, but it needs to be stated explicitly and each one assessed — particularly `path_provider_foundation`, which also shows up as churn in `macos/Flutter/GeneratedPluginRegistrant.swift`. If they can't be avoided, that's a real cost of the dependency and worth weighing against the feature.

**2. Has this been run?** Asking sincerely, because a few things suggest it may not have been exercised much:
- The mouth animation cannot do anything: `animation_engine.dart` hardcodes `mouthOpen => 0`, `mouthWidth => 0` and `mouthRound => 0`.
- The fallback anchor path looks like it produces bad values (details below).
- The expression cross-fade in the chat sidebar is removed by this diff.

If it has been run, on which platforms? Shader support and the TFLite/OpenCV native libraries all vary across Windows/macOS/Linux, and the project has to work on all three.

**3. Is this ready for review, or is it a work in progress?** A meaningful share of the diff is scaffolding for behaviour that isn't wired up — the lip-sync path, the blendshape fields, the eyebrow anchors (see below). That's completely fine as WIP, but it changes what kind of feedback is useful. If it's still in progress, marking it draft would help; I'd rather give you input on the direction than list unfinished pieces.

## Blocking

**Per-frame `setState` for an idle animation.** `animation_viewport.dart` calls `setState(() {})` on every ticker frame, which rebuilds `ShaderBuilder` → `LayoutBuilder` → `CustomPaint` — including a layout pass — 60+ times a second for as long as a chat is open. `_ShaderPainter.shouldRepaint` also returns unconditional `true`. The standard approach is to drive the painter from a `Listenable` via `CustomPainter(repaint: ...)` and not call `setState` at all; the ticker then costs a repaint rather than a rebuild. We've been actively reducing background work lately, so a permanent full-rate rebuild on the main chat screen is the single biggest concern here.

**No way to turn it off.** `AnimationParams.enabled` exists but isn't connected to anything — no Settings toggle, no persisted preference, and no equivalent in the web/mobile UI. Anything user-visible needs a settings surface and web parity (CLAUDE.md), and an always-on animated portrait particularly needs an opt-out for laptops and lower-end GPUs. Defaulting it off initially would also de-risk the rollout.

**Dependency weight.** `face_detection_tflite` brings in `flutter_litert`, `opencv_dart` and `dartcv4`, none of which are in the project today. That's a large amount of native library across three platforms. We removed every sidecar process fairly recently (`docs/design/sidecar-retirement.md`) to keep the install lean, so it'd be good to know the installer size delta before taking this on.

## Other issues

**Unfinished paths.** With the three mouth getters returning `0`, the mouth displacement, cavity shadow and teeth highlight in `displacement.frag` all multiply by zero on every fragment. `FaceLandmarks` extracts and stores 12 blendshapes (`browInnerUp`, `smileLeft`, `jawOpen`, …) that are never read, and the shader declares and receives `u_anchorLeftBrow`/`u_anchorRightBrow` but never uses them. Either wire these up or drop them for now — they're easy to re-add.

**Coordinate mismatch in the fallback path.** `face_detector_service.dart` uses already-normalised defaults (e.g. `const Offset(0.42, 0.65)`), but `live_portrait_avatar.dart` divides every landmark by image width/height. When `mesh == null`, those defaults get divided again and land around `0.001`.

**Detection re-runs per expression change.** `FaceDetectorService` is created per widget, and the widget is keyed `ValueKey(expressionKey)` — so each expression switch disposes the state and re-initialises a TFLite detector plus a fresh `detectFromFile`, alongside `instantiateImageCodec`, all on the UI isolate. Caching landmarks per image path and sharing one detector would avoid this.

**`ui.Image` isn't disposed.** `_loadedImage` is replaced on reload without a `dispose()`.

**Expression cross-fade regression.** The `AnimatedSwitcher` removed at `chat_page.dart:2988` is the 300 ms transition between expression portraits; expression changes will now hard-cut. The themed error state (`AppColors` container + sized icon) is also replaced with a bare `Icon(Icons.person)`.

**Silent error handling.** `catch (e) { return null; }` in `detectFromFile` and `catch (_)` in `_loadImage` — the project asks that errors be logged or surfaced rather than swallowed, since these will be hard to diagnose from a user report.

**Layout and conventions.** `animation_viewport.dart` and `live_portrait_avatar.dart` are widgets under `lib/services/`, and `services.dart` now exports them; they'd belong in `lib/ui/`. The six new files are also missing the AGPL headers every other file carries, and there are no tests.

## Unrelated changes worth reverting

- `create_group_chat_page.dart` and `output_sanitizer_rule_editor.dart`: a stray leading space added to the `onReorderItem:` lines.
- `download_queue_panel.dart` and `hf_model_card.dart`: the comment `// Old axisAlignment: -1 (vertical) per the SDK migration formula.` is deleted. That note was left deliberately to explain the SDK migration and isn't related to this feature.

## Suggested order

If you want to keep momentum: revert the floors file and the unrelated edits, switch the painter to `repaint:`, add the on/off setting (default off, desktop + web), and either finish or remove the mouth/blendshape/brow scaffolding. That would make the rest straightforward to review.

Happy to talk through the `repaint:` change or the settings/web-parity side if that's useful.

---

linux4life1:
Addendum to my review — theming, expanding on a point I made only in passing.

This PR doesn't add hard-coded colours, but it **removes** theming that was there. In `chat_page.dart` the replaced block was:

```dart
errorBuilder: (_, _, _) => Container(
  color: AppColors.resolve(
    context, Colors.black26, Colors.black.withValues(alpha: 0.1),
  ),
  child: Icon(
    Icons.person,
    color: AppColors.iconSecondary(context),
    size: 64,
  ),
),
```

and it becomes `errorBuilder: (_, _, _) => const Icon(Icons.person, size: 64)` — appearing twice in `live_portrait_avatar.dart`, once in the loading branch and once in the `AnimationViewport` fallback.

That drops an `AppColors.resolve` light/dark pair and an `AppColors.iconSecondary(context)` call. The `AppColors.resolve(context, dark, light)` helper exists precisely so a surface can differ between themes, and this was one of the places already doing it correctly. The new version has no container at all and an uncoloured icon, so a broken-image state renders differently — and in light mode, worse — than it does today.

Worth carrying the original `errorBuilder` across rather than simplifying it, and since it now appears twice, lifting it into a single shared widget or helper would avoid the two copies drifting apart later.

---

PR #173, opened 2026-07-30
Add Emotional Voice streaming support + reference docs

dazpants1:
Reworked emotional voice with a simpler approach

### DIFF
--- a/lib/services/chat/chat_service_generation.dart
+++ b/lib/services/chat/chat_service_generation.dart
@@ -1807,6 +1807,7 @@ extension ChatServiceGeneration on ChatService {
             lastMsg.displayText,
             voiceKey: voiceKey,
             messageId: msgId,
+            emotionLabel: _expressionService.currentExpressionLabel,
           );
         }
 
--- a/lib/services/storage/settings/expression_settings.dart
+++ b/lib/services/storage/settings/expression_settings.dart
@@ -29,6 +29,7 @@ class ExpressionSettings with SettingsBase {
   String _expressionFallback = 'neutral'; // 'neutral', 'prime', 'none', 'emoji'
   bool _expressionEmojiBurst = false;
   double _expressionEmojiBurstSize = 28.0; // particle base font size (12–60)
+  bool _emotionalVoiceEnabled = false;
 
   bool get expressionEnabled => _expressionEnabled;
   String get expressionClassificationMode => _expressionClassificationMode;
@@ -37,6 +38,7 @@ class ExpressionSettings with SettingsBase {
   String get expressionFallback => _expressionFallback;
   bool get expressionEmojiBurst => _expressionEmojiBurst;
   double get expressionEmojiBurstSize => _expressionEmojiBurstSize;
+  bool get emotionalVoiceEnabled => _emotionalVoiceEnabled;
 
   void load() {
     _expressionEnabled = prefs?.getBool(k('expression_enabled')) ?? false;
@@ -53,6 +55,8 @@ class ExpressionSettings with SettingsBase {
     _expressionEmojiBurstSize =
         (prefs?.getDouble(k('expression_emoji_burst_size')) ?? 28.0)
             .clamp(12.0, 60.0);
+    _emotionalVoiceEnabled =
+        prefs?.getBool(k('emotional_voice_enabled')) ?? false;
   }
 
   Future<void> setExpressionEnabled(bool value) async {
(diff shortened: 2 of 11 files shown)


linux4life1 (changes requested):
This is a much cleaner submission than the live-portrait PR — AGPL headers are present, the rate table cites its sources, the feature has a real settings toggle and defaults to off, the ElevenLabs style is applied after `activeEngine` resets it (so the user's configured style isn't clobbered), and `_cachedEmoRate` was correctly added to the replay-cache identity. The overall approach is sound.

Requesting changes for a few correctness issues, mostly in `WavResampler`.

## Questions

**1. How was this tested, and against which engines?** `WavResampler` is doing raw byte manipulation on audio, and several of the issues below only show up on specific engine output or specific emotions. Knowing what you exercised (Kokoro / Piper / ElevenLabs / OpenAI, and which emotions) would help scope the review.

**2. Can `WavResampler` get unit tests in this PR?** It's pure Dart with no plugin or platform dependency — you can synthesise a WAV in memory, resample it, and assert frame count, header fields and sample values. Given it's byte-level code, that's where the value is. There are no tests in the diff.

## Correctness

**Non-16/8-bit input silently produces silence.** The interpolation loop only has branches for `bytesPerSample == 2` and `== 1`. For any other depth (24-bit, or 32-bit float) neither branch runs, `dst` stays zero-filled, and the function writes a structurally valid WAV containing pure silence — no exception, no log, and the caller caches it as good audio. Related: `audioFormat` at byte offset 20 is never read, but the output header hardcodes `1` (PCM), so a float-format input would also be mislabelled. Please bail out and return `input` unchanged for anything the loop can't actually handle.

**The `fmt ` chunk is assumed to be at a fixed offset.** The `data` chunk is located by scanning (good), but `channels`, `sampleRate` and `bitsPerSample` are read from hardcoded offsets 22/24/34, which assumes `fmt ` is the first chunk after the RIFF header. A file with a `LIST` or `JUNK` chunk first would be misparsed into nonsense rates. Worth scanning for `fmt ` the same way you scan for `data`.

**The pre-resample file is leaked.** `resample()` writes a new temp file and returns it; every call site then overwrites its reference without deleting the original. In `speak()` the Kokoro/Piper WAV is orphaned, in `generateAudioFile()` the concatenated WAV is orphaned after `_cleanupFiles(wavFiles)` has already run, and the streaming path orphans each chunk. Over a long session that accumulates in the system temp directory.

**Replay uses the current emotion, not the message's.** `message_bubble.dart` passes `chatService.expressionService.currentExpressionLabel`, which is the character's mood *right now*. Replaying an older message therefore re-renders it with whatever the character is feeling at replay time — a line that was delivered sad will play back cheerful if the mood has since changed. Messages already carry their own emotion in metadata; that's the value this should use. The same applies to the `chat_service_generation.dart` call site if generation and playback can be separated in time.

**ElevenLabs cache key doesn't track what actually varies.** For ElevenLabs no resampling happens — the variation comes entirely from `style`. But the cache identity uses `_cachedEmoRate`, and several emotions share a rate while differing in style: `joy` and `excitement` are both `1.06` (style 0.5 vs 0.6); `admiration`, `gratitude`, `love`, `pride` and `approval` are all `1.02` with styles 0.3/0.2/0.4/0.3/0.2. Switching between two such emotions is a cache hit, so the previous style's audio replays. Including the style value (or the emotion label) in the cache identity for ElevenLabs would fix it.

## Parity

`voice_facade.dart` and `voice_routes.dart` accept an `emotion` parameter, but there are no `web_ui/` changes in the diff — nothing on the web side sends it, and the "Emotional Voice" toggle exists only in desktop Settings. So the backend is plumbed but web/mobile users can neither use nor configure the feature. Per CLAUDE.md the settings surface needs a web counterpart in the same body of work, or an explicit deferral from the maintainer.

## Minor

The new `speak(...)` signature is on one long line while the surrounding signatures in that file are wrapped one-parameter-per-line.

---

linux4life1:
Addendum to my review — theming.

The new Emotional Voice switch uses `AppColors.presetColors[4]` for `activeTrackColor`, which resolves to `Color(0xFF8B5CF6)` — purple.

The warm-porch accent standard in CLAUDE.md asks that new chrome (buttons, toggles, borders, spinners, surfaces) use `AppColors.formMasterAccent` or `AppColors.porchAmberOf(context)`, with `AppColors.onChaosAccent` as the foreground on any solid amber fill. Off-palette is allowed only for genuinely semantic colours, and then only with maintainer sign-off plus a trailing `// theme-keep: <reason>` marker. A settings toggle isn't semantic — it's chrome.

To be fair about it: `voice_media_tab.dart` already uses `presetColors[4]` in five places, so this is consistent with its neighbours rather than something you invented, and I'm not asking you to convert the existing five. But new toggles should be landing on amber, otherwise the file keeps accumulating purple.

The rest of this block is good — `AppColors.textSecondary(context)` on both the icon and the label is exactly right, and it's why I don't think the palette was unfamiliar to you.

---

PR #174, opened 2026-07-31
Voice cloning

dazpants1:
Add ZipVoice zero-shot voice cloning (sherpa-onnx)
Summary
Adds ZipVoice, sherpa-onnx's zero-shot voice-cloning model, to local TTS. Any reference voice can now be cloned from a single clip (3–12 s WAV) plus its exact transcript, with quality that passed a live ear test. Runs fully in-process via a persistent worker isolate — no Python, consistent with the sidecar-retirement architecture.
ZipVoice delivers voice cloning with default quality (numSteps: 4) at ~2.3 s per short clone after a ~3.7 s warm-up.
Key changes
- New engine lib/services/tts/sherpa_zipvoice_engine.dart: persistent worker isolate (spawn handshake [ready, modelDir, libDir], serialized job chain, null = shutdown), model auto-download (distill-int8 bundle + vocos vocoder), numSteps: 4, reference-audio + transcript caching.
- Transcript rule: ZipVoice needs the exact transcript of the reference clip. It is read automatically from a sibling .txt (reference.wav → reference.txt). A missing transcript fails with a clear, user-facing error, and the settings panel explains the rule.
- TtsService wiring: zipvoice engine id with single-default-voice fallback, model-ready gating, error surfacing via lastError.
- Settings: zipvoice_tts_reference_audio key with the reference-audio picker; default engine unchanged (kokoro).
- New engine tab "🧩 ZipVoice / Zero-shot" in TTS settings; new EngineHealth row.
Concerns raised before approval — and how each was resolved
1. ONNX Runtime version drift (sherpa bundles ORT 1.27 vs the app's 1.22): verified bit-identical output on both versions across all 14 golden cases — worst-case cosine 0.9999996232659607, worst-case maxAbs 1.0322e-7, identical to all 16 digits. Each run's loaded DLL was verified via GetModuleFileNameW to rule out accidental filename dedup (test/services/embedding/ort_version_probe_test.dart).
2. Shared-sanitizer isolation: an earlier attempt leaked phoneme hacks (a→uh, i→I, …→space) into the shared _sanitizeText. The shared sanitizer is untouched; ZipVoice handles its own text internally. Zero ZipVoice references in the shared sanitize path.
3. Cross-platform ORT layout: the onnxruntime.dll filename collision is Windows-only; macOS ships versioned dylibs and Linux distinct SONAMEs, which coexist by design. On Windows the collision is already handled by the existing force-copy in windows/CMakeLists.txt (~L127–141); built bundles ship ORT 1.27.
4. Engine protocol safety: the worker-isolate handshake/job protocol was validated end-to-end in a standalone prototype before any app wiring (zipvoice_worker_proto.dart, listenable outputs in proto5/), including error isolation and the mandatory-transcript failure mode.
Verification
- flutter analyze clean; full flutter test green (2 pre-existing unrelated Windows path-separator failures in character_authoring_facade_test.dart).
- New unit tests for the engine's pure logic (model layout, presence check, sibling-transcript resolution).
- Windows release build confirmed; manual clone test passed with heart.wav + heart.txt.
Out of scope
- Per-character reference audio (global-only for now).
- Exposing numSteps (8 = slower/higher quality) in the UI — hardcoded to the ear-tested default.

Opencode check of Kokoro bug: not ours. Verified at the VoiceCloning↔Rawhide merge-base (b6498645): all 3 dead entries (jm_beta, em_santa, zm_yibo) exist in the rawhide root too, authored by linux4life1 in 8fc8ea7d5 (2026-05-23). Those voices never worked in the root — they silently fell back to af_heart there as well; it's silent, so it just didn't look broken.
p.join fallout fixed. My app fix broke 2 Windows-only path assertions; I made both separator-agnostic (avatar_gallery_test.dart:53/57, character_repository_test.dart:874). Full suite is now fully green on Windows for the first time: 2745 passed, 12 skipped, 0 failed (same set the owner's Linux CI runs, plus the previously-failing avatar tests now pass on both platforms).


linux4life1 (changes requested):
Technically this is the strongest of the three open PRs. It has AGPL headers, real unit tests for both the new engine and the Kokoro speaker-id change, `EngineHealth` wired up with `expected:` flags, updated goldens, a persistent worker isolate with reference-sample caching, `p.join` used throughout (including the drive-by fix in `avatar_image.dart`), and the Windows `onnxruntime.dll` collision fix carries a clear explanation of why it exists. The `_resolveSid` change is a genuine improvement — a silent fallback becoming a logged one.

Requesting changes on one policy question and a handful of robustness/UX items.

## Policy question — needs a maintainer decision before code review matters

This ships **zero-shot voice cloning from an arbitrary WAV**, with no consent gate, no acknowledgement step, and nothing in the docs about acceptable use. Point any 3–12 second clip of a real person at it and the app reads arbitrary text in their voice.

That's a product decision rather than a code defect, and it should be made deliberately rather than arrive as a side effect of a TTS engine addition. Worth settling: does this need a first-use acknowledgement, wording in the docs/release notes, or any restriction at all? The project already gates The Stoop behind explicit 18+ consent, so there's precedent for treating this kind of thing as a deliberate choice.

## Questions

**1. What was tested, and on which platforms?** Particularly whether the Windows `onnxruntime.dll` fix was verified on a real Windows box with a clean build directory *and* an incremental one — that block exists specifically because the outcome was previously state-dependent, so it needs testing from both starting states. Also whether cloning was run end-to-end (model download → reference pick → transcript → generation) rather than just the unit tests.

**2. Was the ~1GB model download considered against the recent first-launch work?** We just reworked startup specifically so users aren't pushed into large downloads they may not want. The dialog does disclose the size next to the button, which is good; mostly flagging so the interaction is a conscious one.

## Usability

**The transcript requirement makes the feature hard to actually use.** ZipVoice needs the exact transcript of the reference clip, and the only supported way to supply it is to create a `.txt` file with the same basename next to the WAV — outside the app, in a file manager or text editor. The info panel explains this clearly, which helps, but the picker lets you select `reference.wav` and then the feature simply fails until you go and hand-author `reference.txt`.

A `TextField` in that same panel that writes the sibling `.txt` for the selected WAV would close this, and it removes the "select file, get an error" path entirely.

## Robustness

**`generateAudio` throws for a missing transcript but returns `null` for every other failure.** Missing model, missing reference path and a non-existent reference file all report to `EngineHealth` and return `null`; the missing transcript reports *and then* throws `StateError`. `speak()` catches it, but that inconsistency is easy to trip over, and `generateAudioFile` is what the web `/voice` route calls — so with ZipVoice selected as the active engine, a web TTS request depends on a reference file and transcript that can only be configured from the desktop UI. Returning `null` with the same `EngineHealth` report as the other cases would be more predictable.

**The worker's reference cache is keyed on path only.** `_cachedRefPath == refPath` short-circuits the re-read, so replacing the audio at the same path (re-recording `reference.wav`) keeps using the stale samples until the app restarts. A modification-time check alongside the path would cover it.

**The Windows DLL copy indexes into a list by position.** `list(GET sherpa_onnx_windows_bundled_libraries 1 _SHERPA_ONNXRUNTIME_DLL)` assumes `onnxruntime.dll` is the second entry. If sherpa-onnx ever reorders or adds to that list, this silently copies the wrong file and reintroduces exactly the crash the block exists to prevent — and it would fail the same state-dependent way that made the original bug hard to pin down. Matching the list entry by filename would make it self-correcting.

## Parity

There's no web/mobile counterpart: ZipVoice can't be selected as an engine and the reference audio can't be configured from `web_ui/`. Because the active engine is global, a desktop user selecting ZipVoice changes what web TTS does without web users having any way to see or configure it. Per CLAUDE.md that needs either a web settings surface in the same body of work or an explicit maintainer deferral.

## Minor

`numSteps` is a hardcoded `4` described as the ear-test default, with no way to trade speed for quality. Fine for a first cut, but worth a comment noting it's deliberately not exposed yet, or a setting.

---

linux4life1:
Addendum to my review — theming, which I should have included the first time.

The new `_buildZipVoiceSettings` block adds 16 hard-coded colours:

| colour | count |
|---|---|
| `Colors.black26` | 3 |
| `Colors.white38` | 3 |
| `Colors.green` | 3 |
| `Colors.cyan` | 2 |
| `Colors.white12` / `white24` / `white54` / `white` | 1 each |
| `Colors.redAccent` | 1 |

CLAUDE.md is explicit that hard-coded `Color(0xFF...)` and raw `Colors.whiteXX`/`Colors.blackXX` are not allowed in **new or refactored** UI, and that everything should go through `AppColors` — `backgroundOf/cardOf/surfaceOf/surfaceContainerOf(context)`, `textPrimary/Secondary/Tertiary(context)`, `iconSecondary(context)`, `borderOf(context)`, or `AppColors.resolve(context, dark, light)` for a custom accent.

**Why it matters beyond style:** these are fixed values, so they don't respond to light/dark. `Colors.white38` body text and `Colors.black26` panel backgrounds are tuned for a dark surface and will be low-contrast-to-unreadable in light mode. That's an accessibility problem, not just a convention one.

**The question:** is there a reason `Colors.whiteXX` was chosen here over the `AppColors` helpers — something the helpers don't cover, or just following what was already in the file? I ask because the credit is real: the buttons and progress indicators correctly use `AppColors.formMasterAccent` and `AppColors.onChaosAccent`, so the palette clearly wasn't unknown to you. It looks like it was applied to the chrome and not the text/surfaces.

**Context, not an excuse:** `tts_settings_dialog.dart` already contains ~44 raw-colour uses that predate this PR, so the new code is consistent with its surroundings. Nobody expects this PR to fix the existing 44. But the rule applies to new code, and adding 16 more moves that file further from the goal rather than closer.

Separately, `Colors.cyan` for the info icon and the transcript instruction is off-palette chrome. The warm-porch standard wants `AppColors.formMasterAccent` or `AppColors.porchAmberOf(context)`, with `AppColors.onChaosAccent` as the ink on any solid amber fill. A genuinely semantic colour can stay off-palette, but only with maintainer sign-off and a trailing `// theme-keep: <reason>` marker — the "model ready ✓" green arguably qualifies as a status colour; the cyan info panel doesn't.

---

PR #189, opened 2026-08-10
Eval system review: fused reply-facts, tri-state One-Shot, shared judge prefix, dream de-block + hygiene fixes

linux4life1:
## What and why

This branch is the implementation of the full eval-system review (every secondary LLM call: realism judges, one-shot, verification, needs, climax/Afterglow, pockets, posture, scene-time, expression, cast, journal, growth, dreams, ambitions, promises, objectives). The goal was squeezing more performance out of a system whose results are already good — fewer calls, less prompt re-reading, nothing blocking the send path — without changing what any feature does.

**Performance (per turn):**
- **Fused reply-facts call** — climax + pockets + posture share ONE post-generation request when 2+ of them are live. Each feature keeps its own gate (fire = OR, sections = AND), honoring the "Pockets rides no other feature's pass" ruling; transport is the only thing shared. Single-feature turns are byte-identical to before.
- **Shared judge prefix** — the three pre-generation judges (relationship, emotional, narrative) + one-shot now open with a byte-identical intro/dossier/standing/roster/frame block, so local backends KV-fast-forward it instead of re-reading the character sheet three times. Scene-time dispatches last. No eval moved across the pre/post boundary — a guard group pins phase placement.
- **Objective completion check unblocked** — it fired a full LLM request before *every* reply while any quest was open. Now on the quest's own frequency, plus an immediate check when the recent exchange actually mentions the quest (pure mention gate, stemmed, cast/user-name aware).
- **Post-gen parallelized** — the needs eval and the fused reply-facts call run concurrently instead of in sequence.
- **Dream prefetch** — the night's dream was generated INSIDE sendMessage, freezing the user's morning message on a model call. It's now produced at the end of the turn whose clock crossed the night (the rollover is visible by post-generation) and parked; the next send just inserts the finished text. Session-guarded, single-take, mismatch-discarded; same insertion position, journal card, and silent-skip floor.
- **Prompt hygiene** — the needs eval stopped requesting four fields nothing reads; eval sampler settings corrected (no repetition penalty on structured scorecards).

**One-Shot becomes tri-state Auto/On/Off (Auto default)** — Auto resolves to the fused single call on remote backends with confirmed tool support, and the four-call path on local models. Old `true` migrates to On, old `false` to Auto. Desktop pill row + web UI shipped together.

**Correctness fixes found by the review:**
- **Arousal double-apply** — the relationship path parsed `arousal_delta` "best-effort" though its prompt never requests it; a model that volunteered the field got arousal applied twice in one turn (relationship parse + emotional eval), the chip showing only the second value. Both appliers now take a required `applyArousal` ownership flag; the eval whose prompt requests the field is the only one that applies it, in multi-call, one-shot, and the batch arms.
- **Dead hunger verifier rule** — the Director's "hunger delta without eating" check probed the key `hunger` while the eval emits `hunger_delta`; the strict-quote extractor can never match the longer key, so the rule had never fired on real output. The dedicated suite's `anyOf('corrected','accepted')` assertions were green either way. Fixed + strictly pinned.
- **Photo blindness** — `recentExchange` fed needs/climax/pockets `displayText`; the judges read `promptText`. Photo turns (where only promptText carries the "[shared a photo: caption]" marker) scored as the user saying nothing. One-word fix, guarded.
- **`[EvalTraffic]` instrumentation** — one debug line per turn tallying every secondary call (count, prompt chars, output, LLM time), split turn vs background, so regressions in call volume are visible instead of felt.

**Deferred from the review, with reasons** (documented in `.claude/changelog.md`): Journal/Growth structural clone dedup (pure refactor of two live passes; kept out of a PR already carrying behavior changes), guest chime gate batching (`scene_guest_director_test.dart` pins the gates), and a salient-kick minimum window (would break the protected `journal_review_test`/`growth_rings_test` canned-kick fixtures — needs a maintainer call on amending committed tests first).

⚠️ **Heads-up for the `test-integrity` gate:** this branch regenerates the two `character_state_settings` pixel goldens (the tri-state pill row changed that panel's layout) — that will need the `approved-test-change` label. No existing test logic was modified; all 30+ new guards are new files.

## Target branch

- [x] This PR targets the correct branch ([which branch?](https://github.com/linux4life1/front-porch-AI/blob/Rawhide/CONTRIBUTING.md#which-branch-do-i-target)) — features/perf/refactor → Rawhide

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Refactor / cleanup

## How was this tested?

Tested on: Linux (AMD64 sandbox — the same platform CI's golden container uses). Windows/macOS not run locally; CI covers all three, and nothing here touches paths, process spawning, or native libraries.

- [x] `flutter analyze` is clean (`--no-fatal-warnings --no-fatal-infos` → No issues found; `dart fix --dry-run` → nothing to fix)
- [x] `flutter test` passes — full unit suite `--concurrency=4 --exclude-tags golden` → **+3483 green, ~12 skipped, 0 failed**; golden suite `--tags golden` → **+94 green** (including the two regenerated goldens, visually verified)
- [x] **Every new guard proven red-then-green** against the specific thing it guards: fused carrier ignored, dropped required field, drifted rubric, prefix byte flipped, dispatch swapped, Auto branch inverted, session guard removed, single-take removed, postgen kick deleted, promptText reverted, hunger probe reverted, arousal ownership gate dropped (both directions)
- [ ] Ran the app by hand — not possible in this sandbox (no display); the wiring-level suites drive the real `ChatService` through a scripted backend instead (`dream_prefetch_test`, `posture`-style harness)

## Parity

- [x] Web/mobile UI (`web_ui/`) updated to match — the tri-state One-Shot control shipped on both surfaces in the same commit; the remaining changes in this batch have no user-visible surface (timing/transport/internal fixes)
- [x] Realism/Needs behaviour is identical in 1:1 and group chats — the fused call, dream owner rule, and arousal ownership all sit on the shared per-speaker paths; one-shot vs multi-call parity pinned by `one_shot_parity_test` plus the new arousal suite covering both modes

## Checklist

- [x] New files carry the AGPL header
- [x] New/refactored UI uses `AppColors` (the tri-state pill row; no hard-coded colors)
- [x] I did not run `dart format` over whole files (only the newly created test files were formatted)
- [x] I did not edit the version in `pubspec.yaml`
- [x] Errors are logged or surfaced, not silently swallowed (dream prefetch failures log `[Dreams] prefetch skipped` and fall back to the same silent-skip floor the blocking design had)
- [x] My contribution is licensed under AGPL-3.0-or-later and I have the right to submit it

## Screenshots

The tri-state One-Shot row is pinned by the regenerated `character_state_settings.{dark,light}.png` goldens in this diff — they are the before/after for the only visual change in the branch.

https://claude.ai/code/session_0199vEDTKStAAMK4jGFEqzia

---
_Generated by [Claude Code](https://claude.ai/code/session_0199vEDTKStAAMK4jGFEqzia)_


linux4life1:
CI status: the **Guard protected paths** failure is the expected one flagged in the PR description — the only files it names are the two regenerated pixel goldens:

```
modified test/golden/widget/_goldens/sidebar/character_state_settings.dark.png
modified test/golden/widget/_goldens/sidebar/character_state_settings.light.png
```

These changed because the tri-state One-Shot pill row (Auto / On / Off) replaced the old two-state switch in that panel, so the old pixels are genuinely stale. Both PNGs were regenerated in the CI-parity Linux/AMD64 environment and visually verified; the full 94-test golden suite is green with them. Every other flagged-path category passed — the gate itself lists the nine new test suites as "allowed, not blocking."

This check needs the `approved-test-change` label, which the author can't self-apply — nothing else on this PR is waiting on it. I'm watching the remaining checks and will fix anything that comes back red.

---

PR #205, opened 2026-08-20
feat(backend): add Concentrate as a Quick Connect remote API preset

AjayK47:
Front Porch already has a generic OpenAI-compatible client under the hood — LM Studio, OpenRouter, and Nano-GPT are all just the same OpenRouterService pointed at different URLs. Concentrate (https://concentrate.ai — 150+ models across 20+ providers) already worked here through the manual URL field, it just didn't have a one-click Quick Connect button like the others do. This adds that button. No new client code needed since Concentrate speaks the same OpenAI-compatible API the app already knows how to talk to — same shape as the OpenRouter and Nano-GPT entries right above it in remote_api_section.dart. Also updated the web UI to match and rebuilt it (npm run build), and ran flutter analyze / dart fix / dart format — all clean.


linux4life1:
Appreciate the though on adding a quick connect for the provider, however that isn't a path I want to go down. There are literally hundreds of API providers and accepting this PR and merging sets a president of yes I will allow other provider easy connect buttons. It will become a "Button farm" or the UI will have to be refactored to support drop downs etc. also that creates a problem if the API provider shuts down etc. so I'm going to pass on this one. Thank you for the thought though.

PR #222, opened 2026-09-03
Release: promote Rawhide

linux4life1:
## Release: promote Rawhide to main

This PR brings the stable channel up to date with everything that has been
proving itself on the nightly builds. It's a fast-forward promotion — the
tree is Rawhide's, verbatim, with the two stable-facing files (README and
the What's New notes) kept from main.

### What's new for users

- **Calendar birthdays on character and persona cards** — birthdays now
  appear on the card panel and on The Stoop, and verified creators get
  gold/blue checks next to their Discussion posts.
- **Worlds without mandatory weather** — a lorebook-only world can opt out
  of climate entirely, so quiet places stay quiet.
- **Faster, more reliable tool calls** — the eval pipeline no longer stalls
  on empty responses, and the three prefix-sharing judges share one tools
  list, so structured evals land sooner on capable models.

### Major fixes

- **No more quiet data loss** — replacing a portrait no longer wipes Journal
  entries, Growth rings, quests, or RAG embeddings; buried swipes stop
  deleting later entries; and forking a 1:1 chat into a group now copies
  live pockets.
- **Realism chips stay honest** — regenerating a reply keeps its chips and
  needs deltas, and a scoped reprocess only evaluates the needs you ticked.
- **AFK owns the clock** — idle turns stamp the chosen speaker's needs and
  set the story clock before any post-reply eval can drift.

### Verification

CI runs the exact tree that will ship: all workflow files are byte-identical
to Rawhide, guarded stable files match main, no conflict markers, every
workflow YAML parses. Unit + integration suite and E2E smoke tests green
across macOS, Windows, and Linux.


linux4life1:
Closing — superseded by a fresh promotion candidate that guards .claude/changelog.md.

PR #294, opened 2026-09-21
Release: promote Rawhide (v1.4 Toolbox candidate)

linux4life1:
This is the v1.4 Toolbox candidate: Rawhide’s tree on main, with the stable README, in-app What’s New, and changelog pointer kept from main.

What ships is the drawer that finally does work. Waifu Coder runs OpenCode the way Kobold already runs. Search and wiki ride the reply. A second local model can sit beside chat for feelings and journal. Guests can sit on the group couch without taking the house until you Promote them. Image Studio uses Comfy’s own templates. Opening a chat lands on the latest line. Afterglow limp is the first reply, not the whole cooldown.

The long-form release page is Rawhide’s copy (it is not guarded). The in-app dialog is main’s Toolbox notes, including the cuts that landed after the first draft.

CI on this PR is the gate. Do not squash; ancestry has to stay a real merge. Do not fast-forward main until this is green and you say so.


linux4life1:
Closing: re-cut after #295 lands on Rawhide. Guarded docs + faq/user-guide 1.4 stamps already on main/Rawhide — nothing unique left on this tip.

